PR #18, opened 2025-03-12
Item Capabilities fix

jojo7682345:
A fix for gregtech machines not generating items with the proper heat tag.

Fixed by resolving capabilities when items are transfered.

(May contain unused imports)


Pyritie:
Since the other person in GTM said that this could come with a slight performance hit, would it be feasible to check if the item you're resolving is one of the few items that even needs to be resolved? (Basically anything that has heat data.) Or is this not feasible to do from this point in the mixin?

TFG doesn't have a high volume of items going around, so I suspect that the performance hit of always resolving capabilities won't be too big, but hey everything helps

jojo7682345:
I'll look into it, I first tried to only resolve capabilities when the item is transferred to an ItemHandler from another mod, however this required searching the block registry, which I suspect is worse than resolving capabilities every time an item is transferred.

jojo7682345:
I added filtering for TFC items.

Pyritie:
awesome! thanks so much for figuring this all out

jojo7682345:
No problem, this issue was driving me crazy.

Would it be possible for someone else to remove the unused imports, as my IDE is not set up properly, and constantly recompiling to check if an import was required is quite a hassle.

Pyritie:
sure, would you be able to duplicate this over in https://github.com/Pyritie/Core-Modern/tree/0.9 ?  That's where me and some other contributors are actively working on stuff since the lead dev is busy

jojo7682345:
Sure, but that will have to wait until tomorrow.

krossgg:
Wanted to chime in and say that if you're updating to a newer version of GTM (1.5+) - we don't use this method anymore. 

My recommendation is that you try and mixin specifically near `NotifiableItemStackHandler#L145`ish and add the resolution there by getting the stack in the current slot and checking if it's something with the heat cap (only when `simulate == false`), since you really only need it to happen when the recipe outputs are added to the bus.

jojo7682345:
Thank you for the recommendation. However the version of GTM that the modpack uses, hardcodes the simulate flag to false, and I am unable to completely remove that line of code with mixins. Causing the code to be ran twice, without having the ability to check if it is simulated or not. The new version of  handleRecipe does thinks way better. 

Therefore I think the implementation from the first two commits are the best to implement for the current version of the modpack. I did add a new solution based on the input from krossgg, but disabled it for now, as it runs twice for every item produced. It can be enabled once the modpack moves to GTM 1.5 or above (with minor alterations) and the other version can be disabled.

Pyritie:
We're moving to GTM 1.6.4! That's the main "feature" of the 0.9 version we're working on (the 0.9.x branch here, for example)

jojo7682345:
Ok, I have enabled the right fix for GTM 1.6.4, although I can't really test if it actually works, but it should. (If not I have probably made a spelling mistake somewhere)

PR #29, opened 2025-05-01
Render TFC ingots using their mod textures if valid

AirRice:
## What is the new behavior?
Welcome back gold.
TFC ingot piles will now fetch and use their corresponding TFC textures if they exist, only falling back to our current system (GTCEU material colours on top of base universal texture) if the metal texture does not exist.

## Implementation Details
Initially I checked just to see if the metal texture existed in TFC and then fell back to our base texture-based rendering if it didn't. Turns out TFC forces all metals it doesn't recognize into unknown. We thus check if the ingot pile *thinks* it's unknown, then crossreference GT's material registry - If it's 'seen as' unknown but has a material registry, we use our fallback rendering technique (previous behaviour), and otherwise it uses the TFC rendering.

## Outcome
- Ingots with valid TFC metal data AND a corresponding texture will show up as that texture in ingot piles.
- Ingots with no valid TFC metal data that still has GT Material data will use a colour tinted base universal texture (previous behaviour) to be rendered in ingot piles.
- Ingots with valid TFC metal data but NO corresponding texture will show up as a **missing texture**

## Additional Information
Before change           | After change
:-------------------------:|:-------------------------:
Before change|After change

Ingots pictured are:
Rose gold, Silver,
Gold, Copper, Black Steel.

Mixed pile
Mixed pile showing the difference between gold (TFC rendering) and aluminium (TFG rendering) and that they can coexist in one pile.

After merging this all 'metals' defined through TFC data that have a pileable ingot MUST have a corresponding texture or they will start showing up as a missing texture. So far this happens for red alloy and tin alloy, which will probably have textures added soon into the base modpack repository to also fix issues such as them showing up as missing in mold tables. I find this a non-issue due to the very small number of metals we actually add as TFC-compatible, it's all mostly done on the GT side of things.

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/client/TFGClientHelpers.java
+++ b/src/main/java/su/terrafirmagreg/core/client/TFGClientHelpers.java
@@ -11,7 +11,6 @@ import net.minecraftforge.api.distmarker.OnlyIn;
 
 @OnlyIn(Dist.CLIENT)
 public final class TFGClientHelpers {
-
     /**
      * Просто скопированный метод из RenderHelper.java (TFC) + добавленный аргумент для цвета.
      * */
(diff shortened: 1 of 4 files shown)


Pyritie:
BEAUTIFUL!

Out of curiosity, did you happen to notice why the generated piles don't have any shading, now that there's pictures side by side?

Pyritie (dismissed):

AirRice:
@Pyritie Yes actually! it's because the TFC pile rendering has code to shade the ingot piles, but for the TFG core rendering methods we have seem to have copied over what was there except the shading code. The piles don't have shading as a result, but it shouldn't be too bad to bring that over at some point

AirRice:
Actually I'm going to turn this into a draft for now, I forgot to also make this apply to double ingot piles + I think I can get shading working too.

AirRice:
Copper/gold/aluminium mixed pile on left, Titanium pile on right
Newest commit makes GT ingot piles also have proper shading.

Pyritie (approved):

Redeix (approved):

Redeix:
Nice!!

PR #42, opened 2025-05-31
-Added Trowel tool

Redeix:
Core code for modpack #1106 https://github.com/TerraFirmaGreg-Team/Modpack-Modern/pull/1106


src/main/java/su/terrafirmagreg/core/common/data/items/TrowelItem.java
```diff
@@ -0,0 +1,202 @@
...
+
+public class TrowelItem extends Item {
+    public TrowelItem(Properties properties) {
+        super(properties.durability(1026));
+    }
+
+    public static Map<ResourceLocation, ResourceLocation> createBlockMapping() {
+        Map<ResourceLocation, ResourceLocation> map = new HashMap<>();
+
+        String[] sandstone_colors = {
```
Pyritie, line 33: would it be cleaner to use TFC's SandBlockType or Rock.BlockType enums here? or would that be too tricky

Redeix: the only issue with that is that I dont know how I would tie it into the resulting block for each type

Pyritie: what about just turning the enums into strings? then you can do a similar thing, just without making a new array

Pyritie (approved):
perfect, thank you!

PR #43, opened 2025-05-31
Removes GT tool recycling

Pyritie:
## What is the new behavior?
Added mixins to disable arc furnace and macerator tool recycling. The recipes aren't shown in JEI, it's a "special behaviour" these machines have.

## Implementation Details
Normally gregtech calculates how much you'd get back based on the durability, but it uses the default tool crafting recipe as a starting point (such as 3 ingots for a pickaxe) and it doesn't look like that sort of "material value" property is easy to change.

The machines have similar recipes for turbine rotors but we haven't changed those so it's fine to keep them imo.

## Outcome
Can't duplicate materials via tools any more. But also can't recycle them, outside of what you can put in a crucible.

## Additional Information
This does mean that people can't recycle their tools any more... but given that this bug wasn't reported until now, I don't think it's something that will be missed, and our tools are so much cheaper than default anyway.


Redeix:
We could always make recipe results depending on nbt data. So its not like the option is gone if we really wanted it

Pyritie:
yeah, "a way to repair tools" is a more popular request than "a way to recycle tools"

PR #62, opened 2025-07-01
- Added kjs builder methods for double tall plants and smoke emitters.

Redeix:
I want to add some more of these, and possibly allow control over the smoke particles through kubejs as well.

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/common/data/blocks/DoubleDecorativePlantBlock.java
+++ b/src/main/java/su/terrafirmagreg/core/common/data/blocks/DoubleDecorativePlantBlock.java
@@ -1,0 +1,59 @@
+package su.terrafirmagreg.core.common.data.blocks;
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.core.Direction;
+import net.minecraft.world.item.Item;
+import net.minecraft.world.level.BlockGetter;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.LevelAccessor;
+import net.minecraft.world.level.LevelReader;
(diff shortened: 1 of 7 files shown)


src/main/java/su/terrafirmagreg/core/common/data/blocks/DoubleDecorativePlantBlock.java
```diff
@@ -0,0 +1,59 @@
...
+			return belowState.is(this) && belowState.getValue(BlockStateProperties.DOUBLE_BLOCK_HALF) == DoubleBlockHalf.LOWER;
+		}
+	}
+
+	@Override
+	public void onPlace(BlockState state, Level level, BlockPos pos, BlockState oldState, boolean isMoving) {
+		if (state.getValue(BlockStateProperties.DOUBLE_BLOCK_HALF) == DoubleBlockHalf.LOWER) {
+			BlockPos above = pos.above();
+			if (level.getBlockState(above).canBeReplaced()) {
+				level.setBlock(above, this.defaultBlockState().setValue(BlockStateProperties.DOUBLE_BLOCK_HALF, DoubleBlockHalf.UPPER), 3);
```
Pyritie, line 55: wouldn't this let you only place the bottom half, if the top half was obstructed? I think it should prevent it completely, either giving you the item back or just dropping it on the floor where you placed it (I've seen mods do both things)

Redeix: I'll test it again, and remove those loot tables first!

PR #93, opened 2025-08-09
Add Electric Extendo Grip

dimethylene:
Added Electric Extendo Grip
Consumes 8 EU a tick with 1000000 max
Increases block range by 1 more than regular extendo grip
the code is janky but it seems to work

[](https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/1545)


Pyritie:
does it consume EU constantly? or only when it's equipped/interacting with stuff that's far away

dimethylene:
It consumes EU if it is in your main or off hand

dimethylene:
i completely removed any uses of maxUses, since the item will not have durability anyway, and also cleared persistentdata for when not in inventory

i did notice that in survival, blocks cannot be broken if the item is in the main hand, but it works perfectly in the offhand so i'm not to worried about it

Redeix (approved):

dimethylene:
oh wait i didn't realize making a new commit would discard reviews
oh well

Pyritie:
dw about it

gustovafing (approved):

PR #99, opened 2025-08-11
Prospector Rework

Redeix:
Now scans blocks projecting from the players viewport. And logs ore in the chat, and displays an xray particle effect depending on tier. All tag driven. Removed original functionality.


Pyritie (approved):

src/main/java/su/terrafirmagreg/core/client/OreHighlightRenderer.java
```diff
@@ -0,0 +1,60 @@
...
+        if (event.phase != TickEvent.Phase.END) return;
+
+        long now = System.currentTimeMillis();
+
+        synchronized (highlights) {
+            Iterator<Highlight> iterator = highlights.iterator();
+            while (iterator.hasNext()) {
+                Highlight highlight = iterator.next();
+                if (highlight.expireTime < now) {
+                    iterator.remove();
```
Vazde, line 53: Consider using `highlights.removeIf(h -> h.expireTime < now);` instead of the whole iteration thing. It's also better for performance and safety.

I personally tend to avoid mutating collections while iterating them, unless the collection explicitly states that to be ok - not sure how it's typically in Java.

PR #128, opened 2025-09-08
DNA syringes

Redeix:


src/main/java/su/terrafirmagreg/core/config/ServerConfig.java
```diff
@@ -18,6 +22,8 @@
  */
 public final class ServerConfig {
 
+    private static final Logger LOGGER = LogUtils.getLogger();
```
Pyritie, line 25: there's already a logger in TFGCore, why make another one?

Pyritie (dismissed):

PR #132, opened 2025-09-09
GTUtil.canSeeSunClearly rain fix mixin

FiNiTe-Weeb:
## What is the new behavior?
canSeeSunClearly, used in checking if solar boilers should be working, sometimes thinks the sun isnt clearly visible due to GT assuming that if rain is happening somewhere, its happening everywhere where rain can happen (which isn't true in TFC)

shouldnt be any

**Enter your Discord nickname, if your PR is successfully accepted, you will be given the Contributor role**
finite


Redeix (approved):
You may want to cache blockpos.above() since that call can be pretty laggy and running it twice is redundant.
So like: `BlockPos posAbove = blockPos.above();` or whatever

Pyritie (approved):

PR #138, opened 2025-09-16
Feature/cleanroom orbit

BlueBoat29:
cleanroom in orbit

- [x] recipes work in orbit
- [x] recipes work in a cleanroom
- [x] recipes don't work outside either


gustovafing (approved):

Redeix:
I'm wondering if an oxygenated environment should be left out? @Pyritie what do you think?

Pyritie:
like, oxygenated = not clean? I think that would be confusing

PR #153, opened 2025-09-26
Switch ae2 to Cosmic Frontiert fork

ko-lja:
## What is the new behavior?
Cooler and better ae2

## Implementation Details
Switch repo to the gtceu one and change dependency version

## Outcome
Feature: more cool features i guess???

## Additional Information
Will now need to use the -all.jar when building

Very low chance of breaking addons, currently only pcc is affected, but that's being taken care of by @Pyritie


Pyritie:
I'm guessing we'll need to remove ae2 from pakku too? will anything new need to be added there?

ko-lja:
Since it's bundled in core now, yes. I'm pretty sure that's all that needs changed

PR #157, opened 2025-09-29
Adding Deanfission mod in TFG

TomPlop:


Pyritie (approved):
I assume these blocks are just placeholders for now?

TomPlop:
> I assume these blocks are just placeholders for now?

yeah some of them for efficiency and throttle but we can't add them in the reactor anyway.

Pyritie:
diamond blocks are obtainable so it might be worth changing those to something else unused, like netherite blocks

PR #162, opened 2025-09-30
Per dimension wireless terminal consumption

ko-lja:
## What is the new behavior?
Adds kubejs syntax to allow users to tweak extraction/insertion cost for the wireless terminal per-dimension

## Implementation Details
Add TFGServerEvents so it can easily be reloaded for testing,

## Outcome
Blah blah see above

## Additional Information
Usage example:
```js
const $ResourceLocation = Java.loadClass('net.minecraft.resources.ResourceLocation');

TFGServerEvents.dimensionalPowerConsumption(event => {
    event.add("minecraft:the_nether", 50);
    event.add(new $ResourceLocation("minecraft:the_end"), 100);
});
```
Only other thing that needs to be done is to create a texture, model and lang for the wireless card("tfg:wireless_card") in kubejs


src/main/java/su/terrafirmagreg/core/compat/kjs/events/TFGAE2PowerConsumption.java
```diff
@@ -0,0 +1,28 @@
...
+
+import dev.latvian.mods.kubejs.event.EventJS;
+import it.unimi.dsi.fastutil.objects.Object2ObjectOpenHashMap;
+
+import su.terrafirmagreg.core.TFGCore;
+
+public class TFGAE2PowerConsumption extends EventJS {
+    public static final Object2ObjectOpenHashMap<ResourceLocation, Double> powerConsumption = new Object2ObjectOpenHashMap<>();
+
+    public void add(ResourceLocation level, double power) {
```
Pyritie, line 13: wouldn't the api be better if it just accepted a string, and then java turned it into a resource location?

Pyritie (dismissed):

gustovafing (approved):

Pyritie (approved):

PR #163, opened 2025-10-01
fix wireless terminal..

ko-lja:
## What is the new behavior?
fix the wireless terminal only opening when out of range of the access point and not having a wireless card

## Implementation Details
don't invert the range check, as the method we redirect already does that
## Outcome
fix the terminal not opening..


Redeix (approved):
Removing excitement 😔

Pyritie (approved):

PR #167, opened 2025-10-03
Updated particle emitters

Redeix:
Updated particle emitters to be able to create assignable block entities. And cleaned them up a bit since they were kinda cluttered.
- If you want an emitter to form as a block entity you can add `.hasTicker(true)` to the register. BE's have multiple benefits; they can continue to create particles even if a player is out of range, and can have the frequency fully asjustable.
- You can now adjust the frequency of the random ticks with `.emitDelay(int)`. Settings the int to 0 will have particles appear every tick. The higher the number the longer the delay between random ticks. Essentially it creates a range, where a random number is picked in that range. The top end of the range is the sum of `emitDelay + 1`. Non block entities dont really get affected by this. Since the way regular blocks are ticked is controlled by an internal bernoulli distribution.
- You can now set the initial spawning point of the particles instead of them always spawning from the center of the block by adding `.particleBase(x, y, z)` to the register. For example, `.particleBase(0.5, 10.0, 0.5)` spawns the particle in the center of the block on the x/z but 10 blocks up on the y axis.

I will update the wiki tonight.

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/common/data/TFGBlockEntities.java
+++ b/src/main/java/su/terrafirmagreg/core/common/data/TFGBlockEntities.java
@@ -52,7 +52,6 @@ public class TFGBlockEntities {
                     () -> BlockEntityType.Builder.of(ReflectorBlockEntity::new, TFGBlocks.REFLECTOR_BLOCK.get())
                             .build(null));
 
-
     // Shared particle emitter ticker entity.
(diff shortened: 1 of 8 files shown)


Pyritie (approved):
I have no idea what half of this does lmao

BlueBoat29 (approved):
tested it and it works

PR #168, opened 2025-10-04
Adding new components to the fission reactor

TomPlop:


TomPlop:
> imo it would make more sense to use like a wrench to remove the moderators but that isn't a big deal

lol true but who cares

BlueBoat29:
ok then change it :4983pepediamondsword:

Pyritie (approved):

PR #291, opened 2026-02-08
Adds gradle tasks to execute shortcuts and to buildAndLaunch

Ujhik:
## What is the new behavior?
Adds 2 gradle tasks: executeShortcut and buildAndLaunch. 
Adds colors and styles for logging in gradle

It doesn't modify anything about building nor any behaviour of the gradle, simply adds options to execute shortcuts and to buildAndLaunch

image


gustovafing:
I think this can be closed because of the gradle changes merged with datagen.

PR #314, opened 2026-02-22
Mixin to fix /reload

Mqrius:
Wipe categoryMap on reload so EMI loads new values on /reload

Temporary mixin to fix https://github.com/GregTechCEu/GregTech-Modern/pull/4683 until GT does a new release.


Pyritie (approved):

TomPlop (approved):
yeah really smart way to do it I would have done the same good job Marius

PR #320, opened 2026-02-25
Datagen Fixes

### LINKED ISSUE #3078: [💡]: Electronic Trowel
### Was this feature suggested before?

- [x] No, it was not.

### A list of popular features/mods the team has already considered and will not implement.

- [x] I understand :(

### What would you like to be added?

Add an electronic equivalent to a trowel

### Why is this needed?

I'd really like to avoid crafting hundreds of trowels for construction.
Because this is all I've managed to build with one trowel, and in my opinion, that's very little:
Image

I can also offer an option if the trowel can be repaired

Redeix:
- Fixed missing greenhouse casings from datagen
- Fixed wrong item textures for fish roe and proginator cells
- Fixed events bus being missing
- Added firmalife planter support for harvest baskets
- Updated the trowels to use the RNR recipes instead of manual mappings. And made them unbreakable.

Closes https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3078

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/common/data/TFGBlocks.java
+++ b/src/main/java/su/terrafirmagreg/core/common/data/TFGBlocks.java
@@ -731,7 +731,7 @@ public final class TFGBlocks {
     public static BlockEntry<Block>[] createGreenhouseCasings(String tier, List<TagKey<Block>> blockTags, List<TagKey<Item>> itemTags) {
         List<BlockEntry<ConnectedGlassBlock>> casings = new ArrayList<>();
 
-        for (int i = 0; i < 4; i++) {
+        for (int i = 0; i < 5; i++) {
             String blockId = "casings/greenhouse/%s_greenhouse_casing_%s".formatted(tier, i);
             var blockBuilder = TFGCore.REGISTRATE.block(blockId, ConnectedGlassBlock::new)
                     .initialProperties(() -> Blocks.GLASS)
--- a/src/main/java/su/terrafirmagreg/core/common/data/items/TrowelItem.java
+++ b/src/main/java/su/terrafirmagreg/core/common/data/items/TrowelItem.java
@@ -1,19 +1,16 @@
 package su.terrafirmagreg.core.common.data.items;
 
 import java.util.ArrayList;
-import java.util.Arrays;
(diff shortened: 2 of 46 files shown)


src/main/java/su/terrafirmagreg/core/common/data/items/TrowelItem.java
```diff
@@ -110,51 +64,37 @@ public static Map<ResourceLocation, ResourceLocation> createBlockMapping() {
...
         ResourceLocation clickedBlockId = clickedState.getBlock().builtInRegistryHolder().key().location();
 
-        Map<ResourceLocation, ResourceLocation> blockMapping = createBlockMapping();
-
-        // Exception for rnr roads.
-        if (clickedBlockId.toString().equals("rnr:base_course")) {
-            List<ItemStack> validStacks = new ArrayList<>();
-            // Randomly chooses a spot in the hotbar.
+        // RNR road exception. Use BlockModRecipe to find a matching recipe.
+        if (TFGModsResolver.RNR.isLoaded() && clickedBlockId.toString().equals("rnr:base_course")) {
```
Pyritie, line 68: if we have rnr as a dependency, wouldn't it be better to just check against its actual block, instead of block ID string checking?

src/main/java/su/terrafirmagreg/core/common/data/TFGBlocks.java
Pyritie, line 742: I think this might need to be changed too? if the ornate one should also be cutout and not translucent

src/main/java/su/terrafirmagreg/core/common/data/items/TrowelItem.java
Redeix: Honestly I did this because I am thinking about forking the trowel into a separate mod and I wanted to test how to do an optional dependency

src/main/java/su/terrafirmagreg/core/common/data/TFGBlocks.java
Redeix: Nah ornate it also translucent not cutout

Pyritie (approved):

PR #321, opened 2026-02-25
Custom spawn locations

Mqrius:
Don't merge.


src/main/java/su/terrafirmagreg/core/common/data/utils/CustomSpawnSaveHandler.java
```diff
@@ -0,0 +1,47 @@
...
+import net.minecraft.server.level.ServerLevel;
+
+/// Manages a single GlobalPos that represents the servers world spawn
+public class CustomSpawnSaveHandler extends SaveHandler {
+
+    private GlobalPos data = GlobalPos.of(ServerLevel.OVERWORLD, BlockPos.ZERO);
+
+    @Override
+    public void loadData(CompoundTag compoundTag) {
+        GlobalPos.CODEC.parse(NbtOps.INSTANCE, compoundTag).result().ifPresent(globalPos -> data = globalPos);
```
Mqrius: saveData saves to compoundTag "spawn", but loadData tries to load from root rather than the compoundTag. This silently fails, and the code will end up doing the whole finding valid spawn chunks as if it's a first boot. This _usually_ works but by accident and it can fail in interesting ways.

Try this
```java
  @Override
  public void loadData(CompoundTag compoundTag) {
      GlobalPos.CODEC.parse(NbtOps.INSTANCE, compoundTag.get("spawn")).result().ifPresent(globalPos -> data = globalPos);
  }
```

src/main/java/su/terrafirmagreg/core/common/data/utils/CustomSpawnSaveHandler.java
```diff
@@ -0,0 +1,47 @@
...
+        });
+    }
+
+    public static GlobalPos getSpawnPos(ServerLevel level) {
+        return read(level).data;
+    }
+
+    public static void setSpawnPos(ServerLevel level, GlobalPos globalPos) {
+        CustomSpawnSaveHandler handler = read(level);
+        handler.data = globalPos;
```
Mqrius, line 34: This is where `handler.setDirty()` goes; you set a field dirty any time you change it from its saved value. That only happens in this method

src/main/java/su/terrafirmagreg/core/common/data/utils/CustomSpawnSaveHandler.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public static CustomSpawnSaveHandler read(ServerLevel level) {
+        return (CustomSpawnSaveHandler) read(level.getDataStorage(), CustomSpawnSaveHandler::new, "tfg_spawn_position_data");
+    }
+
+    //This probably doesn't need to always be true, I just don't know enough about it
+    @Override
+    public boolean isDirty() {
+        return true;
```
Mqrius: See above, don't need this bit, just mark dirty whenever you change the data from what is stored

src/main/java/su/terrafirmagreg/core/mixins/common/tfc/ForgeEventHandlerMixin.java
```diff
@@ -44,4 +66,94 @@ public class ForgeEventHandlerMixin {
         // Otherwise do as usual
         return original.call(level, pos, newState);
     }
+
+    /**
+     * @author
+     * @reason It was becoming too complicated to do as a normal mixin.
```
Mqrius, line 87: If you want this more targeted rather than an overwrite, it would look something like this:

```java
    @Redirect(method = "onCreateWorldSpawn", at = @At(value = "INVOKE", target = "Lnet/dries007/tfc/world/ChunkGeneratorExtension;findSpawnBiome(Lnet/minecraft/util/RandomSource;)Lnet/minecraft/core/BlockPos;", remap = true))
    private static BlockPos tfg$findSpawnBiome(ChunkGeneratorExtension extension, RandomSource random, @Local(name = "level") ServerLevel level) {
        RegionGenerator regionGen = new RegionGenerator(extension.settings(), random);

        var condition = CustomSpawnHelper.getFromConfig();

        boolean climateMatch = false;
        int seedTicker = 0;

        BlockPos pos = null;
        ChunkPos chunkPos;
        while (!climateMatch) {

            pos = CustomSpawnHelper.findSpawnBiome(condition.spawnCenterX(), condition.spawnCenterZ(), extension.settings().spawnDistance() * condition.spawnRadiusMultiplier(), random, extension);
            chunkPos = new ChunkPos(pos);
            Region.Point regionPoint = regionGen.getOrCreateRegionPoint(Units.blockToGrid(chunkPos.getMinBlockX()), Units.blockToGrid(chunkPos.getMinBlockZ()));

            System.out.println("Testing chunkPos " + chunkPos.getWorldPosition());
            System.out.println(regionPoint.temperature);
            System.out.println(regionPoint.rainfall);
            if (CustomSpawnHelper.testWithinRanges(regionPoint.temperature, regionPoint.rainfall, condition)) {
                climateMatch = true;
            } else {
                ++seedTicker;
                random = new XoroshiroRandomSource(level.getSeed() + seedTicker);
            }
        }

        return pos;
    }

    @Inject(method = "onCreateWorldSpawn", at = @At(value = "INVOKE", target = "Lnet/minecraftforge/event/level/LevelEvent$CreateSpawnPosition;setCanceled(Z)V"))
    private static void tfg$saveSpawnPos(LevelEvent.CreateSpawnPosition event, CallbackInfo ci, @Local(name = "level") ServerLevel level, @Local(name = "levelData") ServerLevelData leveldata) {
        var condition = CustomSpawnHelper.getFromConfig();

        BlockPos spawnPos = new BlockPos(leveldata.getXSpawn(), leveldata.getYSpawn(), leveldata.getZSpawn());
        GlobalPos globalSpawnPos = GlobalPos.of(ServerLevel.OVERWORLD, spawnPos);

        if (condition.dimension() == ServerLevel.OVERWORLD) {
            CustomSpawnSaveHandler.setSpawnPos(level, globalSpawnPos);
        } else if (condition.dimension() == ServerLevel.NETHER) {
            CustomSpawnSaveHandler.setSpawnPos(level, CustomSpawnHelper.BENEATH_PLACEHOLDER);
        } else if (condition.dimension() == Planet.MARS) {
            CustomSpawnSaveHandler.setSpawnPos(level, CustomSpawnHelper.MARS_PLACEHOLDER);
        }

        CustomSpawnHelper.resetConfigValue();
    }
```
Specifically injecting on the spawn biome search, and at the end to do the save. It's more obvious what has changed this way though the overwrite is fine too

src/main/java/su/terrafirmagreg/core/common/data/utils/CustomSpawnHelper.java
```diff
@@ -0,0 +1,217 @@
...
+    //Adapted from TFC code, but with more config
+    public static BlockPos findSpawnBiome(int spawnCenterX, int spawnCenterZ, int spawnRadius, RandomSource random, ChunkGeneratorExtension extension) {
+        int step = Math.max(1, spawnRadius / 256);
+        int centerX = QuartPos.fromBlock(spawnCenterX);
+        int centerZ = QuartPos.fromBlock(spawnCenterZ);
+        int maxRadius = QuartPos.fromBlock(spawnRadius);
+        BlockPos found = null;
+        int count = 0;
+
+        for (int radius = maxRadius; radius <= maxRadius; radius += step) {
```
Mqrius, line 102: The outer loop never iterates, because radius starts at maxRadius. Is that intentional? Can either remove the outer loop, or start close in and loop outwards if that's what you want

src/main/java/su/terrafirmagreg/core/mixins/common/tfc/ForgeEventHandlerMixin.java
```diff
@@ -44,4 +66,94 @@ public class ForgeEventHandlerMixin {
...
+            if (!foundExactSpawn) {
+                LOGGER.warn("Unable to find a suitable spawn location!");
+            }
+
+            if (level.getServer().getWorldData().worldGenOptions().generateBonusChest()) {
+                LOGGER.warn("No bonus chest for you, you cheaty cheater!");
+            }
+
+            if (condition.dimension() == ServerLevel.OVERWORLD) {
+                CustomSpawnSaveHandler.setSpawnPos(level, globalSpawnPos);
```
Mqrius, line 167: If globalSpawnPos is null then this will cause a NPE when you getSpawnPos.

TFC doesn't have this issue because they do ` levelData.setSpawn(chunkPos.getWorldPosition().offset(8, generator.getSpawnHeight(level), 8), 0.0F);` earlier as a rough first pass. Could do a similar fallback like that for globalSpawnPos by initiating it to that instead of null.

src/main/java/su/terrafirmagreg/core/common/data/utils/CustomSpawnHelper.java
BlueBoat29: I double checked the method in TFC and they also have this the same. Likely it is supposed to be -maxRadius, but its working atm so I don't want to touch it

Pyritie: I could send them a PR with the fix and see what they think

BlueBoat29: https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/52e7bc6c385242d991fe4d9f0bbe6e7731a89906/src/main/java/net/dries007/tfc/world/biome/BiomeSourceExtension.java#L69

src/main/java/su/terrafirmagreg/core/mixins/common/tfc/ForgeEventHandlerMixin.java
BlueBoat29: I actually already use that exact line of code to avoid there not being a worldSpawn if shit hits the fan, so I can use that same blockPos to init globalSpawnPos

src/main/java/su/terrafirmagreg/core/common/data/utils/CustomSpawnHelper.java
Pyritie: https://github.com/TerraFirmaCraft/TerraFirmaCraft/pull/3476

Pyritie: tfc just merged it so I guess `-maxRadius` is correct!

Mqrius: It's really not :sob: it should be zero instead. negative maxRadius makes no sense, you'll be doing double the work for no reason.

Actually the whole outer loop makes no sense, because the inner loop is scanning over the given surface already. So now you're actually doing O(n²) work for no reason.

Mqrius: (Also it's not a radius, it's a max horizontal dimension. Radius implies a circle)

BlueBoat29: Sounds like you need to go in and do some fixes on TFC :trollface:

BlueBoat29:
lgtm :trollface:

src/main/java/su/terrafirmagreg/core/common/data/utils/CustomSpawnHelper.java
```diff
@@ -0,0 +1,228 @@
...
+    /// @param id string used for mapping
+    /// @param spawnCenterX int block pos estimate on the X axis
+    /// @param spawnCenterZ int block pos estimate on the Z axis
+    /// @param spawnRadiusMultiplier int multiplier on default radius
+    /// @param temperatureRange inclusive range to check for the temperature
+    /// @param rainfallRange inclusive range to check for the rainfall
+    /// @param dimension level that this spawn occurs in
+    /// @param difficulty translatable component that displays the difficulty
+    public record CustomSpawnCondition(
+            String id,
```
Redeix, line 233: How come it uses coordinates and climate data? Wouldn't it make sense to just use climate data so it has compatibility with custom world scales?

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
```diff
@@ -65,6 +120,89 @@ public static void onLevelUnload(LevelEvent.Unload event) {
...
+                BlockPos validSpawn = null;
+
+                int chunkSearchRadius = 32;
+
+                Predicate<Holder<Biome>> lushCavePredicate = testBiome -> testBiome.is(ResourceLocation.fromNamespaceAndPath(TFGCore.MOD_ID, "nether/lush_hollow"));
+
+                int count = 0;
+                while (Objects.isNull(validSpawn)) {
+                    ChunkPos chunkPos = new ChunkPos(random.nextInt(chunkSearchRadius * 2) - chunkSearchRadius, random.nextInt(chunkSearchRadius * 2) - chunkSearchRadius);
+                    System.out.println("ChunkPos: " + chunkPos);
```
gustovafing: Remove this

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
```diff
@@ -65,6 +120,89 @@ public static void onLevelUnload(LevelEvent.Unload event) {
...
+                            }
+                        }
+
+                    }
+
+                    if (count >= 25) {
+                        //validSpawn = BlockPos.ZERO;
+                        chunkSearchRadius = chunkSearchRadius * 2;
+                        count = -1;
+                        System.out.println("No Valid Spawn :(, trying search radius of " + chunkSearchRadius);
```
gustovafing: Remove this

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
```diff
@@ -65,6 +120,89 @@ public static void onLevelUnload(LevelEvent.Unload event) {
...
+                    }
+
+                    count++;
+                }
+
+                System.out.println("Found valid spawn point: " + validSpawn);
```
gustovafing: Remove this

src/main/java/su/terrafirmagreg/core/common/data/utils/CustomSpawnHelper.java
```diff
@@ -0,0 +1,228 @@
...
+
+import su.terrafirmagreg.core.config.TFGConfig;
+
+public class CustomSpawnHelper {
+
+    public static final GlobalPos BENEATH_PLACEHOLDER = GlobalPos.of(ServerLevel.NETHER, BlockPos.ZERO);
+    public static final GlobalPos MARS_PLACEHOLDER = GlobalPos.of(Planet.MARS, BlockPos.ZERO);
+
+    public static void respawnTeleporter(ServerPlayer player, ServerLevel targetLevel, GlobalPos worldSpawn) {
+        System.out.println("attempting to spawn player at: " + worldSpawn);
```
gustovafing: Remove this

gustovafing (changes requested):

src/main/java/su/terrafirmagreg/core/common/data/utils/CustomSpawnHelper.java
Redeix: Everyone hates logs until we ask for latest.txt 😔

src/main/java/su/terrafirmagreg/core/common/data/utils/CustomSpawnHelper.java
BlueBoat29: The idea is that the coordinates center the search so it finds it faster, but I could look into multiplying it by the custom scale factor :clueless:

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
BlueBoat29: might keep this one

gustovafing: You should be using the actual logger, not this

BlueBoat29: yeah you right

src/main/java/su/terrafirmagreg/core/common/data/utils/CustomSpawnHelper.java
Redeix: That makes sense. I would definitely look into making it compatible with world scales. It should be as simple as multiplying the set value but the config value :clueless:

Redeix (approved):

gustovafing (approved):

PR #324, opened 2026-02-26
Fix offhand lag

### LINKED ISSUE #3193: [🐛]: Couple of seconds freeze when I take something in the offhand

- [x] No, it was not.

### Computer OS

arch linux 6.12.73-1-lts

### Modpack Version Used

0.11.24

### Minecraft Environment

Forge Server

### New Worlds

Yes

There is no any logs in the moment of freeze

### Expected Behavior

It shouldn't freeze

### Actual Behavior

It freezes when I take any item in the offhand. Not always, but really often.

### Steps to Reproduce

Join any world, take any item in the offhand. Torch, shield, wooden cane e.t.c.

### Additional Information

I use nvidia-open drivers, but another person that plays with me on intel's GPU has the same behavior.

Mqrius:
## What is the new behavior?
Fixes https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3193

## Implementation Details
Prevent AAAParticles hand renderer mixins from applying. We don't use the hand rendering for anything so we don't need it.

What's actually happening is kinda funny.
AAAParticles checks with its registries if anything wants to emit particles when you have something in a hand.
It tries to account for lag, so it checks, hey, when did we last do this? And if it's longer ago it runs through its registries more often to maybe spawn more particles.
However, this check doesn't actually run if there's nothing in your (off)hand... so the moment you put something in your offhand, minutes later, it's like "wow we haven't spawned any particles in several minutes! Let's run through my registries a bazillion times cos maybe we need to spawn a bazillion particles!"

Of course TFG doesn't have any particles that spawn from your hands, but that doesn't really matter, it still creates a bazillion particle spawning contexts.

It's also clearer now why some people experience this a lot and some people don't. If you usually have something in your offhand, you won't experience this lag. And also after the first time it happens it's fine for a while.
You can probably also experience it in your mainhand if you usually keep that empty

## Outcome
AAA Particles doesn't touch hand rendering in any way

## Additional Information
AAA Particles is only used for the sandworm mod, and only for the dirt splash when it enters or leaves the Earth. As far as I can tell there's no other places that use it.
Dirt splash still works cos that's a different code path.

Also reported upstream: https://github.com/ChloePrime/AAAParticles/issues/67

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/mixins/TFGMixinPlugin.java
+++ b/src/main/java/su/terrafirmagreg/core/mixins/TFGMixinPlugin.java
@@ -18,21 +18,43 @@ public class TFGMixinPlugin implements IMixinConfigPlugin {
     private static final String SPECIES_MIXIN_JSON = "species.mixins.json";
     private static final String SPECIES_BLOCK_ENTITY_TYPE_MIXIN = "com.ninni.species.mixin.BlockEntityTypeMixin";
 
+    private static final String AAAPARTICLES_MIXIN_JSON = "aaa_particles.mixins.json";
+    private static final String AAAPARTICLES_GAME_RENDERER_MIXIN = "mod.chloeprime.aaaparticles.mixin.client.MixinGameRenderer";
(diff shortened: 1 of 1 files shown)


Pyritie:
goddamn sandworm mod

Pyritie (approved):

PR #325, opened 2026-02-26
Add /tfg debug dump_recipes, dump_recipes_detailed, dump_pipe_recipes

Mqrius:
## What is the new behavior?
Ingame command to write recipes to disk.

`dump_recipes` writes all recipe IDs in a big json
`dump_recipes_details` writes recipes with inputs and outputs
`dump_pipe_recipes` writes all recipes that use a pipe as input (mostly here as an example to adapt in the code if you need anything else specific)

It sorts the recipes alphabetically by ID so it's easy to diff against if we do a refactor or update.

It's gated at access level 2 which is anyone who's op. It doesn't affect the world in any way though so it's more to hide the debug command.


Pyritie (approved):
damn, that was fast

Mqrius:
It was already done when I mentioned it to you on discord!

PR #326, opened 2026-02-28
Add Artisan Table and casings

TomPlop:


src/generated/resources/data/tfg/tags/blocks/casings.json
```diff
@@ -23,6 +23,7 @@
     "tfg:casings/machine_casing_vacuum_engine_intake",
     "tfg:casings/machine_casing_ultraviolet",
     "tfg:casings/machine_casing_egh",
+    "tfg:casings/sterling_silver_casing2",
```
Mqrius: The other casings use an underscore, like `casing_2`. Let's keep it consistent or Gus will get mad!

src/main/java/su/terrafirmagreg/core/common/data/recipes/ArtisanType.java
```diff
@@ -168,6 +172,21 @@ private static ResourceLocation textureLocation(String name) {
             0.8f,
             2.0f);
 
+    private static final Supplier<Item> opticalBorosilicateBlank = () -> ForgeRegistries.ITEMS
```
Mqrius: static final variables are conventionally written in all caps with underscores, so OPTICAL_BOROSILICATE_BLANK

run/client/kubejs/assets/jade/lang/en_us.json
```diff
@@ -1,7 +0,0 @@
-{
-  "config.jade.plugin_tfc.charger": "",
-  "config.jade.plugin_tfc.soarer": "",
-  "config.jade.plugin_tfc.snatcher": "",
-  "config.jade.plugin_tfc.surfer": "",
-  "config.jade.plugin_tfc.axolotl": ""
-}
```
Mqrius, line 7: Don't commit this deletion

Mqrius (approved):

src/main/java/su/terrafirmagreg/core/common/data/recipes/ArtisanType.java
```diff
@@ -168,6 +172,21 @@ private static ResourceLocation textureLocation(String name) {
...
+            .getValue(TFGCore.id("optical_borosilicate_blank")).asItem();
+    public static final ArtisanType GLASS_LENS = new ArtisanType(
+            "optical_borosilicate",
+            new ItemStack(opticalBorosilicateBlank.get().getDefaultInstance().getItem(), 1),
+            new ItemStack(TFCItems.POWDERS.get(Powder.FLUX).get().getDefaultInstance().getItem(), 8),
```
Mqrius: We don't really need the getDefaultInstance here since `new ItemStack` makes a new stack anyway, and the borosilicate supplier is also not doing much except making the line shorter. So these two lines could just be
```java
new ItemStack(ForgeRegistries.ITEMS.getValue(TFGCore.id("optical_borosilicate_blank"))),
new ItemStack(TFCItems.POWDERS.get(Powder.FLUX).get(), 8),
```
Though it doesn't really break anything either.

src/generated/resources/data/tfg/tags/blocks/casings.json
Pyritie: do we have anything actually using this casing? I'd avoid creating it for now since we might end up recoloring it for a different material.

PR #327, opened 2026-02-28
Add Artisan Table

TomPlop:


Pyritie (approved):

src/main/java/su/terrafirmagreg/core/common/data/recipes/ArtisanType.java
```diff
@@ -4,6 +4,7 @@
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Objects;
+import java.util.function.Supplier;
```
Mqrius, line 7: This one is no longer used, try building, the spotless wil remove it

Mqrius (approved):

src/main/java/su/terrafirmagreg/core/common/data/recipes/ArtisanType.java
Pyritie: I got it

PR #329, opened 2026-03-01
Feature/gt converter fe pull

DrEthan77:
## What is the new behavior?
Adds the ability for all Converters to pull FE from the Portable Energy Interface

## Implementation Details
This limits which FE containers Converters can pull from based on the Block ID, which is somewhat hacky but workable without changing Create Addition's side.

## Outcome
Resolves the issue below
[#2460](https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/2460)

If the Block ID of the Portable Energy Interface is changed, this breaks

**Enter your Discord nickname, if your PR is successfully accepted, you will be given the Contributor role**
DrEthan

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
+++ b/src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
@@ -12,9 +12,10 @@ import com.gregtechceu.gtceu.api.machine.trait.NotifiableEnergyContainer;
 import com.gregtechceu.gtceu.common.machine.trait.ConverterTrait;
 
 import net.minecraft.core.Direction;
+import net.minecraft.core.registries.BuiltInRegistries;
 import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.state.BlockState;
-import net.minecraftforge.registries.ForgeRegistries;
 
 @Mixin(ConverterTrait.class)
 public abstract class ConverterTraitMixin extends NotifiableEnergyContainer {
@@ -24,7 +25,7 @@ public abstract class ConverterTraitMixin extends NotifiableEnergyContainer {
     }
 
     @Inject(method = "serverTick", at = @At(value = "INVOKE", target = "Lcom/gregtechceu/gtceu/api/machine/trait/NotifiableEnergyContainer;serverTick()V"), remap = false)
-    private void tryFeExtract(CallbackInfo ci) {
+    private void tfg$tryFeExtract(CallbackInfo ci) {
         var frontFacing = machine.getFrontFacing();
         var thisEnergyContainer = GTCapabilityHelper.getForgeEnergy(machine.getLevel(),
                 machine.getPos(), null);
@@ -32,10 +33,10 @@ public abstract class ConverterTraitMixin extends NotifiableEnergyContainer {
             if (d == frontFacing)
                 continue;
             BlockState state = machine.getLevel().getBlockState(machine.getPos().relative(d));
-            ResourceLocation id = ForgeRegistries.BLOCKS.getKey(state.getBlock());
+            Block PORTABLE_ENERGY_INTERFACE = BuiltInRegistries.BLOCK.get(ResourceLocation.tryBuild("createaddition", "portable_energy_interface"));
             var targetEnergyContainer = GTCapabilityHelper.getForgeEnergy(machine.getLevel(),
                     machine.getPos().relative(d), null);
-            if (targetEnergyContainer != null && targetEnergyContainer.canExtract() && id.getPath().contains("portable_energy_interface")) {
+            if (targetEnergyContainer != null && targetEnergyContainer.canExtract() && state.is(PORTABLE_ENERGY_INTERFACE)) {
                 int energyExtracted = targetEnergyContainer.extractEnergy(
                         thisEnergyContainer.receiveEnergy(
                                 FeCompat.toFe(
--- a/src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
+++ b/src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
@@ -35,7 +35,6 @@ public abstract class ConverterTraitMixin extends NotifiableEnergyContainer {
             if (d == frontFacing)
                 continue;
             BlockState state = machine.getLevel().getBlockState(machine.getPos().relative(d));
-            Block PORTABLE_ENERGY_INTERFACE = BuiltInRegistries.BLOCK.get(ResourceLocation.tryBuild("createaddition", "portable_energy_interface"));
             var targetEnergyContainer = GTCapabilityHelper.getForgeEnergy(machine.getLevel(),
                     machine.getPos().relative(d), null);
             if (targetEnergyContainer != null && targetEnergyContainer.canExtract() && state.is(PORTABLE_ENERGY_INTERFACE)) {
--- a/src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
+++ b/src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
@@ -19,6 +19,8 @@ import net.minecraft.world.level.block.state.BlockState;
 
 @Mixin(ConverterTrait.class)
 public abstract class ConverterTraitMixin extends NotifiableEnergyContainer {
+    @Unique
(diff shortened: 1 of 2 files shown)


Redeix (dismissed):

Mqrius:
I haven't looked at the FE code yet so I can't tell if the mixin actually works as expected. Some other notes tho:

I think abstract classes don't need a constructor. Since this is a mixin, the mixin class as such doesn't really exist after compilation so the constructor is dead code that really only needs to exist if the ide/compiler otherwise complain. Ah but I guess they let you access `machine` like that? Would also work with an `@Shadow` but this works too.

If you rename your injected method to tfg$tryFeExtract then it'll be easier to recognize in stacktraces and spark profiles a year down the line. Since this method is injected into the target class, you lose the package information on it.

src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
```diff
@@ -0,0 +1,49 @@
...
+        var thisEnergyContainer = GTCapabilityHelper.getForgeEnergy(machine.getLevel(),
+                machine.getPos(), null);
+        for (Direction d : Direction.values()) {
+            if (d == frontFacing)
+                continue;
+            BlockState state = machine.getLevel().getBlockState(machine.getPos().relative(d));
+            ResourceLocation id = ForgeRegistries.BLOCKS.getKey(state.getBlock());
+            var targetEnergyContainer = GTCapabilityHelper.getForgeEnergy(machine.getLevel(),
+                    machine.getPos().relative(d), null);
+            if (targetEnergyContainer != null && targetEnergyContainer.canExtract() && id.getPath().contains("portable_energy_interface")) {
```
Mqrius: Instead of string comparison, it's better to use `state.is(Block)` or `state.is(ResourceLocation)`, it's faster and less error prone.

src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
```diff
@@ -0,0 +1,49 @@
...
+                int energyExtracted = targetEnergyContainer.extractEnergy(
+                        thisEnergyContainer.receiveEnergy(
```
Mqrius, line 42: S~~ince this is simulated to figure out how much you can receive and extract, so you can just try receiving max-int and it'll tell you exactly how much fits. No need to calculate energycapacity minus energystored. It's a pretty common pattern.~~ With the conversions maybe it's not ideal because you run the risk of overflows so this is fine.

src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
```diff
@@ -0,0 +1,49 @@
...
+                machine.getPos(), null);
+        for (Direction d : Direction.values()) {
+            if (d == frontFacing)
+                continue;
+            BlockState state = machine.getLevel().getBlockState(machine.getPos().relative(d));
+            ResourceLocation id = ForgeRegistries.BLOCKS.getKey(state.getBlock());
+            var targetEnergyContainer = GTCapabilityHelper.getForgeEnergy(machine.getLevel(),
+                    machine.getPos().relative(d), null);
+            if (targetEnergyContainer != null && targetEnergyContainer.canExtract() && id.getPath().contains("portable_energy_interface")) {
+                int energyExtracted = targetEnergyContainer.extractEnergy(
```
Mqrius, line 41: If you use FeCompat.extractEu it'll do conversion automatically. Right now I think you don't do conversion with the ratio, which happens to work because I think our ratio is 1. In general if you use insertEu and extractEu it'll be clearer in the code that we're always handling EU amounts

DrEthan77:
> I haven't looked at the FE code yet so I can't tell if the mixin actually works as expected. Some other notes tho:
> 
> I think abstract classes don't need a constructor. Since this is a mixin, the mixin class as such doesn't really exist after compilation so the constructor is dead code that really only needs to exist if the ide/compiler otherwise complain. Ah but I guess they let you access `machine` like that? Would also work with an `@Shadow` but this works too.
> 
> If you rename your injected method to tfg$tryFeExtract then it'll be easier to recognize in stacktraces and spark profiles a year down the line. Since this method is injected into the target class, you lose the package information on it.

I did build and it "works for me!"™.

I have to have the constructor because I'm extending that other class, I did try shadowing the machine var first but it kept complaining, that was before I just built so maybe it would work, I just trusted what intellij was telling me xD.

I'll rename when I get a chance.

src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
DrEthan77: I tried without doing any ratio and when extracting and adding I ended up with more FE than I started with so it's definitely not 1 to 1. I ended up working just in FE because I'm working with FEcontainers for both machines which made it easier since both can use the FE numbers.

Mqrius: Fair enough!

src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
DrEthan77: I tried to do this, but it is saying "Cannot resolve method 'is(ResourceLocation)'", state is blockstate correct?

Mqrius: Yeah state is BlockState.
My bad, ResourceLocation isn't an option, but there's these:
```
is(TagKey<Block>)
is(TagKey<Block>, Predicate<BlockStateBase>)
is(HolderSet<Block>)
is(Block)
```

I guess you didn't end up adding Create Additions as a dependency? Cos if you did then you can do
```java
state.is(CABlocks.PORTABLE_ENERGY_INTERFACE.get())
```

Without the dependency you could do something like
```java
Block PORTABLE_ENERGY_INTERFACE = BuiltInRegistries.BLOCK.get(new ResourceLocation("createaddition", "portable_energy_interface"));
state.is(PORTABLE_ENERGY_INTERFACE);
```
And that should still work at runtime as long as Create Additions is loaded.

Pyritie: it's currently uncraftable because it doesn't work with the converters. With this PR, we can make them craftable again

Mqrius: If you did end up adding CA as a dependency, you can commit it as part of the PR

src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
```diff
@@ -0,0 +1,50 @@
...
+
+import net.minecraft.core.Direction;
+import net.minecraft.core.registries.BuiltInRegistries;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.block.Block;
+import net.minecraft.world.level.block.state.BlockState;
+
+@Mixin(ConverterTrait.class)
+public abstract class ConverterTraitMixin extends NotifiableEnergyContainer {
+
```
Mqrius, line 24:
```suggestion
    @Unique
    private static final Block PORTABLE_ENERGY_INTERFACE = BuiltInRegistries.BLOCK.get(ResourceLocation.fromNamespaceAndPath("createaddition", "portable_energy_interface"));

```

You can extract this bit to a static since it's always the same.
It's better not to use `tryBuild` here because it would catch an informative error and return null, which would then cause a (less informative) null pointer error.

src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/ConverterTraitMixin.java
```diff
@@ -0,0 +1,50 @@
...
+    @Inject(method = "serverTick", at = @At(value = "INVOKE", target = "Lcom/gregtechceu/gtceu/api/machine/trait/NotifiableEnergyContainer;serverTick()V"), remap = false)
+    private void tfg$tryFeExtract(CallbackInfo ci) {
+        var frontFacing = machine.getFrontFacing();
+        var thisEnergyContainer = GTCapabilityHelper.getForgeEnergy(machine.getLevel(),
+                machine.getPos(), null);
+        for (Direction d : Direction.values()) {
+            if (d == frontFacing)
+                continue;
+            BlockState state = machine.getLevel().getBlockState(machine.getPos().relative(d));
+            Block PORTABLE_ENERGY_INTERFACE = BuiltInRegistries.BLOCK.get(ResourceLocation.tryBuild("createaddition", "portable_energy_interface"));
```
Mqrius:
```suggestion
```
See above

Mqrius (approved):

Redeix (approved):

PR #332, opened 2026-03-01
bugfix: desync between ingot mold temperature in server and client

### LINKED ISSUE #2593: [🐛]: [TFC Bug] Full ingot molds usually can't be rightclicked if they solidified from molten

- [x] No, it was not.

### Version modpack used

0.11.10, 0.11.13, 0.11.15

### Environment

Singleplayer

### New Worlds

Yes

### Expected Behavior

Right click extracts ingot when the ingot mold is full and solid

### Actual Behavior

If the ingot mold has been liquid and then turns solid, rightclicking often doesn't do anything and picks up the entire mold instead. After placing it in an inventory slot again, rightclicking works again.

### Steps to Reproduce

Make sure you're in survival. Right clicking ingot molds doesn't work in creative. (side-quest bug!)

Fill up ingot molds with hot metal in the crucible, I recommend tin because it has a low melting temperature.
Let the mold with molten metal solidify. Try to rightclick it; it picks up the mold. Put it down somewhere, and rightclick again. Now it works.

https://github.com/user-attachments/assets/7ef0bcc6-fca0-4d76-8418-e0d7a99c60bb

### Additional Information

Ingot molds that get filled without ever being liquid don't seem to have this problem.

Putting this bug here because it's a TFC 1.20 issue and they probably won't care. Doesn't seem to happen in 1.21.

Ujhik:
## What is the new behavior?
Fixes not being able to unmold a solidified metal mod the first time by fixing a desync of ingot mold temperature between client and server.
This has the side effect of making molds cool down slower when there is metal in them, supposedly this is the intended behaviour.

closes TerraFirmaGreg-Team/Modpack-Modern#2593


Pyritie (dismissed):

src/main/java/su/terrafirmagreg/core/mixins/client/tfc/MoldCapabilityMixin.java
```diff
@@ -0,0 +1,54 @@
...
+    @Shadow
+    private boolean initialized;
+
+    /**
+     * @author Ujhik
+     * @reason To fix ingot molds having different heat values on server and client because of a dummy initial value on heatCapacity messing up with the forge Capability sync system generating inconsistencies. By initializing it to the correct value, we ensure the temperature calculations stay consistent between client and server
+     */
+    @Inject(method = "<init>(Lnet/minecraft/world/item/ItemStack;ILnet/minecraft/tags/TagKey;)V", at = @At("TAIL"), remap = false)
+    private void onInit(ItemStack stack, int capacity, TagKey<Fluid> fluidTag, CallbackInfo ci) {
+        final CompoundTag tankTag = stack.getTagElement("tank");
```
Mqrius: This mixin seems to do essentially the same as this.load(). Since it's on the init, I think initialized is always false so load() will even call updateHeatCapacity(). Or is there a reason to do things "manually" instead?

Mqrius (dismissed):

src/main/resources/tfg.mixins.json
```diff
@@ -4,6 +4,7 @@
         "client.minecraft.IBlockModelGeneratorsInvoker",
         "client.tfc.ClientForgeEventHandlerMixin",
         "client.tfc.DrinkableMixin",
+        "client.tfc.MoldCapabilityMixin",
```
Mqrius: Why is this in the Client package, but loaded in common? Client is for mixins that run only clientside, and they go in the client mixins at the bottom. I think this one should run in common though? (Not that you're the only one doing this wrong lol)

src/main/java/su/terrafirmagreg/core/mixins/client/tfc/MoldCapabilityMixin.java
Ujhik: Good catch ty!, you are right, no good reason, I just got the working version and was so tired i didn't look how to simplify it

src/main/resources/tfg.mixins.json
Ujhik: you are right, mistake, fixed!

PR #334, opened 2026-03-05
Add customStill to some materials to change the textures

TomPlop:


src/main/java/su/terrafirmagreg/core/common/CommonProxy.java
```diff
@@ -90,6 +91,7 @@ public void onRegisterMaterialRegistry(final MaterialRegistryEvent event) {
     public void onPostRegisterMaterials(final PostMaterialEvent event) {
         TFGHelpers.isMaterialRegistrationFinished = true;
         TFGMaterialHandler.postInit();
+        TFGModifyMaterials.modify(event);
```
Mqrius: No need to pass along `event` if you're not using it for anything

src/main/java/su/terrafirmagreg/core/common/data/TFGModifyMaterials.java
```diff
@@ -0,0 +1,26 @@
...
+        enableCustomStill("iron_iii_chloride");
+        enableCustomStill("soldering_alloy");
+        enableCustomStill("tin");
+    }
+
+    private static void enableCustomStill(String materialName) {
+        Material material = GTMaterials.get(materialName);
+
+        FluidBuilder builder = material.getProperty(PropertyKey.FLUID).getQueuedBuilder(FluidStorageKeys.LIQUID);
+        if (builder != null) {
```
Mqrius, line 26: _Usually_ it's good to guard against null, but honestly in a bunch of cases you would actually prefer to crash instead. If there's something wrong, right now it just silently passes and you never find out. There's an argument to be made to crash hard here instead if builder == null, because that shouldn't happen!

Mqrius (approved):

PR #335, opened 2026-03-05
Tool tag prefixes and other MV/HV rework stuff

Pyritie:


src/main/java/su/terrafirmagreg/core/common/data/TFGModifyMaterials.java
Pyritie, line 1: @TomPlop why not pass in the material directly like `GTMaterials.Iron3Chloride` or whatever?

TomPlop: No reasons just didn't think of it

PR #338, opened 2026-03-06
Remove unnecessary EventBusSubscriber annotation from ClientProxy

sqcode06:
## What is the new behavior?
This PR removes the `@Mod.EventBusSubscriber` annotation from `ClientProxy`.

`ClientProxy` is already instantiated only on the client through `DistExecutor` in `TFGCore`, and it is already registered to the mod event bus through the proxy setup in `CommonProxy`. Because of that, the automatic event bus subscriber annotation is redundant.

Leaving the annotation in place causes Forge to try to automatically load `ClientProxy` during mod subscriber discovery on a dedicated server. Since `ClientProxy` references client-only classes such as particle provider registration code, this can lead to dedicated server startup failure with `NoClassDefFoundError` for client-side classes.

This PR ensures that `ClientProxy` is only loaded through the intended client-only proxy path and not through automatic subscriber scanning.

## Outcome
- Fixes a server-side classloading issue caused by client-only references inside `ClientProxy`.

Discord nickname: sqcode

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/client/ClientProxy.java
+++ b/src/main/java/su/terrafirmagreg/core/client/ClientProxy.java
@@ -1,7 +1,6 @@
 package su.terrafirmagreg.core.client;
 
 import static earth.terrarium.adastra.client.forge.AdAstraClientForge.ITEM_RENDERERS;
-import static su.terrafirmagreg.core.common.data.TFGEntities.*;
 
 import java.util.function.BiConsumer;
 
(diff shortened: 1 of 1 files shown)


Pyritie (dismissed):
I still don't fully understand why this works when my own build had the exact same change and didn't work for tom, but if it works it works I guess

Mqrius (approved):

Pyritie (approved):

PR #340, opened 2026-03-07
Backported leopard seal from tfc 1.21

Ujhik:
## What is the new behavior?
A backported leopard seal from tfc 1.21


src/main/java/su/terrafirmagreg/core/common/data/entities/ai/amphibian/AmphibiousStrollToPoi.java
```diff
@@ -0,0 +1,46 @@
...
+public class AmphibiousStrollToPoi {
+    public AmphibiousStrollToPoi() {
+    }
+
+    public static BehaviorControl<PathfinderMob> create(MemoryModuleType<GlobalPos> poiPosMemory, Function<LivingEntity, Float> speedModifier, int closeEnoughDist, int maxDistFromPoi) {
+        MutableLong mutablelong = new MutableLong(0L);
+        return BehaviorBuilder.create(
+                (m) -> m.group(m.registered(MemoryModuleType.WALK_TARGET),
+                        m.present(poiPosMemory)).apply(
+                                m, (walkTarget, pos) -> (level, mob, p_258853_) -> {
```
Pyritie, line 30: what is `p_258853_`?

Ujhik: 🤷 No idea, copypasted that file from TFC, its that way in their code:
image

Pyritie (approved):

src/main/java/su/terrafirmagreg/core/common/data/entities/ai/amphibian/AmphibiousStrollToPoi.java
Mqrius: p_258853_ is the starting time of a behavior. In some other places eg TFCSnifferAi it's called pGameTime.
So here it would walk to a new destination every 80 time units, if I'm reading it right.

Mqrius:
I don't have the brain to look today, will try to get it in tomorrow!

Pyritie:
no rush!

src/main/java/su/terrafirmagreg/core/common/data/entities/ai/amphibian/AmphibiousSetWalkTargetAwayFrom.java
```diff
@@ -0,0 +1,104 @@
...
+import net.minecraft.world.entity.PathfinderMob;
+import net.minecraft.world.entity.ai.behavior.OneShot;
+import net.minecraft.world.entity.ai.behavior.declarative.BehaviorBuilder;
+import net.minecraft.world.entity.ai.memory.MemoryModuleType;
+import net.minecraft.world.entity.ai.memory.WalkTarget;
+import net.minecraft.world.entity.ai.util.AirAndWaterRandomPos;
+import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.phys.Vec3;
+
+import java.util.Optional;
```
Mqrius, line 24: Please run spotlessApply or build tasks before the final PR submission, spotless moves a lot of stuff around.

src/main/java/su/terrafirmagreg/core/common/data/TFGItems.java
```diff
@@ -83,6 +83,7 @@ public static void init() {
     public static final ItemEntry<ForgeSpawnEggItem> SNIFFER_SPAWN_EGG = registerSpawnEgg(TFGEntities.SNIFFER, 11285007, 4829025);
     public static final ItemEntry<ForgeSpawnEggItem> WRAPTOR_SPAWN_EGG = registerSpawnEgg(TFGEntities.WRAPTOR, 15767516, 4829025);
     public static final ItemEntry<ForgeSpawnEggItem> SURFER_SPAWN_EGG = registerSpawnEgg(TFGEntities.SURFER, 7644045, 12824430);
+    public static final ItemEntry<ForgeSpawnEggItem> LEOPARD_SEAL_EGG = registerSpawnEgg(TFGEntities.LEOPARD_SEAL, 0xffffff, 0xffffff);
```
Mqrius, line 86: Give the egg some matching colors! It's backgroundColor highlightColor

src/main/resources/assets/tfg/sounds.json
```diff
@@ -0,0 +1,39 @@
...
+        "sounds": [{
+            "name": "minecraft:entity.wolf.hurt",
+            "type": "event"
+        }],
+        "subtitle": "subtitles.entity.tfc.seal.hurt"
+    },
+    "entity.seal.step": {
+        "sounds": [{
+            "volume": 0.5,
+            "name": "minecraft:entity/fish/flop1",
```
Mqrius, line 27: This looks like a file path instead of a package path. Is that intentional? I see it's that way in TFC 1.21 too but...

src/main/java/su/terrafirmagreg/core/common/data/entities/animals/TFCLeopardSeal/TFCLeopardSeal.java
```diff
@@ -0,0 +1,93 @@
+/* Originally from [TerraFirmaCraft] (https://github.com/TerraFirmaCraft/TerraFirmaCraft)
+ * Licensed under the EUPL, Version 1.2.
+ * You may obtain a copy of the License at:
+ * https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ */
+
+package su.terrafirmagreg.core.common.data.entities.animals.TFCLeopardSeal;
```
Mqrius, line 7: Make packages and folders lowercase only

src/main/java/su/terrafirmagreg/core/common/data/entities/ai/amphibian/AmphibiousSetWalkTargetAwayFrom.java
```diff
@@ -0,0 +1,104 @@
...
+                                    if (Helpers.isFluid(state.getFluidState(), FluidTags.WATER)) {
+                                        bestWaterEscapePos = getVectorFartherFromChaser(escapeFromPos, bestWaterEscapePos, destination);
+                                    } else {
+                                        final BlockState belowState = level.getBlockState(pos.below());
+                                        if (!belowState.isAir() && !Helpers.isFluid(belowState.getFluidState(), FluidTags.WATER)) {
+                                            bestLandEscapePos = getVectorFartherFromChaser(escapeFromPos, bestLandEscapePos, destination);
+                                        }
+                                    }
+                                }
+                                if (Helpers.isEntity(attacker, TFCTags.Entities.OCEAN_PREDATORS) && isTargetPosValidForEscape(currentPos, escapeFromPos, bestLandEscapePos)) {
```
Mqrius, line 73: Setting the pathToTarget happens every iteration of the loop. This isn't your fault, TFC 1.21 does that too, and it's just as wasteful (can send them an issue or PR if you want!)

`Helpers.isEntity(attacker, TFCTags.Entities.OCEAN_PREDATORS)` should be a static calculated before the loop.
`isTargetPosValidForEscape` should be inside the loop as a check before updating the best escape pos.
`pathToTarget.set` should only occur once at the end of the loop.

It's not the worst performance wise since being attacked doesn't happen that often, but it bugs me lol

Ujhik:
I close it to merge this with the bison and remove the seal lootTable faster without having to juggle 2 branches

Mqrius (commented):
Oh I had some comments pending. A bunch of it is actually TFC's fault, up to you if you want to brush that up.

src/main/java/su/terrafirmagreg/core/common/data/entities/ai/amphibian/AmphibiousSetWalkTargetAwayFrom.java
Ujhik: Sure. Doesn't It run when I build too? I think this may have to do with me having autoimport activated in IntelliJ and its making a mess, I'll have to disable it sadly

src/main/java/su/terrafirmagreg/core/common/data/TFGItems.java
Ujhik: Sure!

src/main/java/su/terrafirmagreg/core/common/data/entities/ai/amphibian/AmphibiousSetWalkTargetAwayFrom.java
Mqrius: It does run when you build, yes, through the spotlessApply dependency somewhere. Maybe autoimport runs after the spotlessApply? If you can fix the order it'll work fine

src/main/resources/assets/tfg/sounds.json
Ujhik: It looks like both formats are valid, it seems the dots reference sound events and the slashes sound files withe their paths I think. I just tested "entity.seal.step" with /playsound and its working nice, and I see other references with slashes in the tfc sounds, for example the bison ones:
image

Mqrius: Huh fair enough!

src/main/java/su/terrafirmagreg/core/common/data/entities/ai/amphibian/AmphibiousSetWalkTargetAwayFrom.java
Ujhik: You are right, didn't bother to read the inner workings. If I change this I think I will have a hard time testing its working right, would prefer that they fix it yep

src/main/java/su/terrafirmagreg/core/common/data/entities/ai/amphibian/AmphibiousSetWalkTargetAwayFrom.java
Ujhik: Confirmed Its the autoimport just ran spetless and the autoimport instantly messed it up after.... I'll disable it to avoid this, ty. In fact If I try to fix it by hand the autoimport messes it up in real time. Sad having to disable because its such a nice feature

Mqrius: My intellij sometimes still does autoimports but I'm not sure. Sometimes when I'm pasting bits of code?

src/main/resources/assets/tfg/sounds.json
Pyritie: This is correct -- `entity.bison.ambient` is the sound ID, `tfc:entity/deer/say1` points to a filepath

src/main/java/su/terrafirmagreg/core/common/data/entities/ai/amphibian/AmphibiousSetWalkTargetAwayFrom.java
Ujhik: image
You probably have the "insert imports on paste"

I have also the add imports on the fly and optimize them. Probably the optimize its the problem changing them after I build. I'll try to adjust it to spotless conf or if I can't removing it

PR #341, opened 2026-03-08
Improve Supports performance

Mqrius:
Right now any neighbor update (like grass growing) will cause 14 thousand block checks to see if any of them are a horizontal support block.

This PR caches support blocks and has efficient access.

The cache is initially populated when necessary via a ~~hacky~~ elegant high speed low level direct access to a chunk's palette data. It then persists until level unload. The cache is updated when supports are placed or removed. It doesn't get saved to SavedData because the chunk access is sufficiently fast.

Removed Zippity's half finished attempt from last year.

Should save about 3ms per 100 chunks loaded, depending on grass growth.


Redeix (dismissed):
Ngl my eyes kinda glazed over for this one. But I think I sort of understand. Wondering if this is worth mentioning to TFC

Mqrius:
The high level picture is that I'm building a map of ChunkPos -> support blocks. Then if a dirt block wants to know if it's supported, it can just check the map for nearby chunks, see if any support blocks are in range, and answer that way.

The implementation details are mostly about how you keep the map in sync with breaking and adding supports (that's the HorizontalSupportBlockMixin), and how you populate the map in the first place (that's the chunk loading stuff)

TFC might not care because their support max range is much smaller, they only scan 400 blocks per grass growth; we were scanning 14k blocks.

Redeix:
Oh that makes sense. I think I had some of this backwards in my head. I should get a new rubber duck

Pyritie (approved):
if it doesn't work, I'm sure our players will let us know pretty quickly lol

src/main/java/su/terrafirmagreg/core/common/perf/SupportCache.java
```diff
@@ -0,0 +1,244 @@
...
+import net.minecraft.world.level.chunk.PalettedContainer;
+
+import it.unimi.dsi.fastutil.ints.IntArraySet;
+import it.unimi.dsi.fastutil.ints.IntSet;
+
+/**
+ * Per-level cache of support block AABBs.
+ * Replaces Support.isSupported()'s brute-force 14k block scan with a few AABB check.
+ */
+public class SupportCache {
```
gustovafing, line 29: Why don't you use a forge SaveData class instead, it attaches an object to each level instance so you don't need to manage the cache lifecycle yourself. 

You can look at the interplanetary logistics network class, just make the save method empty since there is no persistent data.

gustovafing (changes requested):

src/main/java/su/terrafirmagreg/core/common/perf/SupportCache.java
Mqrius: I considered SavedData, I know how it works, but for people coming in with old worlds with no saved data I still need to populate the cache either way. The current method is actually really fast since it shorts circuits immediately for every chunk section that doesn't have a support in its palette, which is most of them.

Given that it's so fast, I opted not to save the data outside of sessions at all. This will automatically fix stuff like create contraptions moving supports and such, I basically have to worry less about the cache being out of sync with reality.

Mqrius: Or do you not mean persisting the data across sessions? I might have to look into SaveData, does sound cleaner.

gustovafing: yeah you don't need to actually save anything

gustovafing: It's called SavedData but it can also just be used for any data that needs to be attached to the level lifecycle

Mqrius: Yeah sounds cleaner, I'll rewrite it in the coming days, but for now this works okay if we wanna do a release soon

PR #345, opened 2026-03-08
Feature/backport tfc animals leopard seal bison

Ujhik:
## What is the new behavior?
Backports TFC:

- Leopard Seal
- Bison


Pyritie:
the meats should go in kjs

Pyritie (changes requested):
meats and loot tables should to be moved to core, otherwise looks good!

Pyritie (dismissed):

src/main/java/su/terrafirmagreg/core/common/data/entities/ai/amphibian/AmphibiousSetWalkTargetAwayFrom.java
```diff
@@ -0,0 +1,106 @@
...
+                                    } else {
+                                        final BlockState belowState = level.getBlockState(pos.below());
+                                        if (!belowState.isAir() && !Helpers.isFluid(belowState.getFluidState(), FluidTags.WATER)) {
+                                            bestLandEscapePos = getVectorFartherFromChaser(escapeFromPos, bestLandEscapePos, destination);
+                                        }
+                                    }
+                                }
+                            }
+
+                            if (isOceanPredator && isTargetPosValidForEscape(currentPos, escapeFromPos, bestLandEscapePos)) {
```
Mqrius: isTargetPosValidForEscape is checked outside the loop now, so if the "best" target pos turns out not to be valid, then you don't get the second best pos but it'll fall back to the `else` where it's not even checked if the target pos is valid. The `isTargetPosValidForEscape` call should be inside the loop before overwriting bestLandEscapePos

Mqrius: I guess all it does is check whether its direction is away from the predator, so nothing will crash for invalid positions, but yea

Ujhik: Isn't It called anyway in the last loop call so its effectively doing the same as TFC code but only executing once instead of all the loop calls?

Mqrius (approved):

Pyritie (approved):

PR #346, opened 2026-03-09
Boiler rework

TomPlop:


run/client/kubejs/assets/jade/lang/en_us.json
Pyritie, line 1: put this file back pls

TomPlop: Ah I don't get why this one always get deleted

Mqrius: I think it's when you run gradle clean that it deletes the run folder maybe

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/TFGLargeBoilerMachine.java
Pyritie, line 1: could you just subclass the existing large boiler machine class instead of needing to copy and paste all of its code? or is every method different

TomPlop: It's honestly way too different

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/TFGMultiMachines.java
```diff
@@ -1017,5 +1025,147 @@ private static net.minecraft.world.level.block.state.BlockState orientedBlockSta
                 }
             })
             .register();
+
+    public static final MultiblockMachineDefinition TFG_LARGE_BOILER_BRONZE = REGISTRATE
```
Pyritie: don't need the TFG_ prefix because you're already in the TFG namespace

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/TFGLargeBoilerMachine.java
Pyritie: fair enough

Pyritie:
I think you need to merge dev into `feature/large_boiler` too, I see all the bison/seal stuff

Pyritie (approved):

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/TFGLargeBoilerMachine.java
```diff
@@ -0,0 +1,521 @@
...
+                    currentTemperature = Math.max(effectiveMaxTemp, currentTemperature - getCoolDownRate() * 10);
+                }
+            }
+        } else if (currentTemperature > 0) {
+            currentTemperature -= getCoolDownRate();
+        }
+
+        if (isFormed() && getOffsetTimer() % TICKS_PER_STEAM_GENERATION == 0) {
+
+            tryDrainBoostFluid();
```
Mqrius, line 250: This will drain boost fluid even when the boiler is not working. Is that intentional? (It stops when the temperature reaches 0 because then it unsubscribes from ticks)

TomPlop: Yeah because the Boilers run the recipe to increase in temperature but they output steam when their temperature is over 100K

So even if there is no more recipes running the temperature is still over 100 while it goes down and will still produce steam

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/TFGLargeBoilerMachine.java
```diff
@@ -0,0 +1,521 @@
...
+                        }
+                    }
+                }
+            }
+        }
+        updateSteamSubscription();
+    }
+
+    @Nullable
+    private BoosterFluid getBestAvailableBooster() {
```
Mqrius: getBestAvailableBooster is called very often and is somewhat expensive to call. It seems possible to just cache the BoosterFluid along with the Tank that holds it, always try to drain from that Tank, and only refresh the cache if the drain fails.

TomPlop: I was a bit afraid of that yeah I'm not so sure how to do it but I can try to look it up

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/TFGLargeBoilerMachine.java
```diff
@@ -0,0 +1,521 @@
...
+        List<IRecipeHandler<?>> inputTanks = new ArrayList<>();
+        inputTanks.addAll(getCapabilitiesFlat(IO.IN, FluidRecipeCapability.CAP));
+        inputTanks.addAll(getCapabilitiesFlat(IO.BOTH, FluidRecipeCapability.CAP));
+
+        BoosterFluid bestBooster = null;
+        for (BoosterFluid booster : getCompatibleBoosters()) {
```
Mqrius: This iterates over all tanks multiple times (as many times as there are boosters). It's better to do the logic the other way around; iterate only over tanks, check what fluid it holds and how much, and if that's a compatible booster and it's enough, store it. For a small optimization you can short-circuit if you've found a tank containing the best possible booster, because you can guarantee that no further inputTank will get better than that

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/TFGLargeBoilerMachine.java
TomPlop: I will just make a cache that gets reset less often because that sounds easier

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/LargeSteamTurbine.java
```diff
@@ -0,0 +1,292 @@
...
+    public static final int MIN_DURABILITY_TO_WARN = 10;
+
+    private final long BASE_EU_OUTPUT;
+    @Getter
+    private final int tier;
+
+    public LargeSteamTurbine(IMachineBlockEntity holder, int tier) {
+        super(holder);
+        this.tier = tier;
+        this.BASE_EU_OUTPUT = 2048;
```
Mqrius: GTValues.V[tier+1] or GTValues.V[tier] * 4

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/TFGLargeBoilerMachine.java
Mqrius: A "cache" in this case is just a class field that holds a tank reference, and one that holds a fluid reference. Then you can make a small helper function to drain from it (with a simulate boolean), and the helper function can update the cache fields it the drain doesn't work. It's some refactoring work but it'll end up cleaner in the end

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/TFGLargeBoilerMachine.java
```diff
@@ -0,0 +1,521 @@
...
+            if (getOffsetTimer() % 20 == 0) {
+                getRecipeLogic().refreshDurationForTemperature();
+            }
+
+            // Amount of steam output
+            double baseDrainExact = (double) currentTemperature * throttle * TICKS_PER_STEAM_GENERATION /
+                    (ConfigHolder.INSTANCE.machines.largeBoilers.steamPerWater * 100.0);
+
+            // Only for amount of water consummed
+            final int WATER_THRESHOLD = 800; // From that temperature water is getting more consumed
```
Mqrius, line 262: Best to extract the `800` into a class static constant because it's used in multiple places

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/TFGMultiMachines.java
```diff
@@ -1017,5 +1025,147 @@ private static net.minecraft.world.level.block.state.BlockState orientedBlockSta
...
+            .modelProperty(GTMachineModelProperties.RECIPE_LOGIC_STATUS, RecipeLogic.Status.IDLE)
+            .model(GTMachineModels.createWorkableCasingMachineModel(
+                            GTCEu.id("block/casings/solid/machine_casing_bronze_plated_bricks"),
+                            GTCEu.id("block/multiblock/generator/large_bronze_boiler"))
+                    .andThen(b -> b.addDynamicRenderer(
+                            () -> DynamicRenderHelper.makeBoilerPartRender(
+                                    BoilerFireboxType.BRONZE_FIREBOX, GTBlocks.CASING_BRONZE_BRICKS))))
+            .tooltips(
+                    Component.translatable("gtceu.multiblock.large_boiler.max_temperature", 754, 480),
+                    Component.translatable("gtceu.multiblock.large_boiler.heat_time_tooltip", 480 / 1 / 20),
```
Mqrius, line 1071: This heat time should be higher, right? Since max temp is 800, not 480. Or is this for "operating temperature"?

TomPlop: Every values aren't finals

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/TFGLargeBoilerMachine.java
TomPlop: Well I don't know if I want to tweak them honestly I may go with different stuffs for Water and RecipeDuration as it may depends of the rest of the balance

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/TFGLargeBoilerMachine.java
TomPlop: So I do a list of all my Boosters and order them by bonus then check all the Tank to see which has the best fluids in them?

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/LargeSteamTurbine.java
TomPlop: A specific reason to do that rather that directly using the number?

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/TFGLargeBoilerMachine.java
```diff
@@ -317,15 +317,30 @@ private BoosterFluid computeBestAvailableBooster() {
...
         inputTanks.addAll(getCapabilitiesFlat(IO.BOTH, FluidRecipeCapability.CAP));
 
+        // Order the booster by temperatureBonus so we can stop searching if we find the best one
+        List<BoosterFluid> compatibleBoosters = getCompatibleBoosters().stream()
+                .sorted(Comparator.comparingInt(BoosterFluid::temperatureBonus).reversed())
+                .toList();
+        if (compatibleBoosters.isEmpty())
+            return null;
+
+        int bestBonusPossible = compatibleBoosters.get(0).temperatureBonus();
```
Mqrius: compatibleBoosters and bestBonusPossible should be class final fields, initiated in the constructor, since they never change per instance

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/LargeSteamTurbine.java
Mqrius: Yeah, in general it's bad practice to use raw numbers in your code for a few different reasons. Numbers by themselves aren't "self-documenting", it doesn't tell you what they mean. GTValues.V[tier]*4 tells you exactly what it means; 4 Amps for whatever tier machine you're building.
In this case it also automatically works if anything changes, like if you create this machine at different tiers, the code will still work - you have `tier` as a parameter after all.

Mqrius (approved):

PR #350, opened 2026-03-10
Feature/large boiler

TomPlop:
more work on the large boiler
add an event to copy the nbt of the tool when crafting it


src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/multiblock/electric/TFGLargeBoilerMachine.java
```diff
@@ -505,17 +516,16 @@ public void setCurrentThrottle(int currentThrottle) {
...
             if (current <= THRESHOLD)
                 return 1.0;
 
-            double degreesAboveThreshold = current - THRESHOLD;
-            double reduction = (degreesAboveThreshold / 100.0) * REDUCTION_PER_100_DEGREES;
-
-            // Minimum
-            return Math.max(0.1, 1.0 - reduction);
+            double t = (current - THRESHOLD) / 1000.0;
+            // log : monte vite au début, ralentit progressivement vers le cap
```
Pyritie: english pls

src/main/java/su/terrafirmagreg/core/common/data/tfgt/TFGTRepairHelper.java
```diff
@@ -0,0 +1,54 @@
...
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.event.entity.player.PlayerEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+
+import su.terrafirmagreg.core.TFGCore;
+import su.terrafirmagreg.core.compat.gtceu.TFGTagPrefix;
+
+@Mod.EventBusSubscriber(modid = TFGCore.MOD_ID)
+public class TFGTRepairHelper {
```
Pyritie, line 17: why not put this in TFGHelpers? it's already a class we use a lot in the modpack, and it's exposed to kjs so you don't need to do any `Java.loadClass`

src/main/java/su/terrafirmagreg/core/compat/gtceu/TFGTagPrefix.java
```diff
@@ -318,6 +320,17 @@ public final class TFGTagPrefix {
                 .generateItem(true)
                 .generationCondition(hasOreProperty.and(mat -> mat.hasFlag(TFGMaterialFlags.GENERATE_DUSTY_ORES)));
 
+        repairKit = new TagPrefix("repair_kit", true)
+                .idPattern("repair_kit_%s")
+                .defaultTagPath("repair_kit_materials/%s")
+                .itemTable(() -> GTMaterialItems.MATERIAL_ITEMS)
+                .unformattedTagPath("repair_kit_materials")
+                .materialAmount(GTValues.M * 4)
```
Pyritie: this means every repair kit is worth 4 ingots, I don't think that's what you want

TomPlop: yeah I forgot out tools were cheap as fuck

src/main/java/su/terrafirmagreg/core/common/data/tfgt/TFGTRepairHelper.java
TomPlop: I actually don't call it in KubeJS it just runs when I do a repairing

Mqrius (approved):

src/main/java/su/terrafirmagreg/core/common/data/tfgt/TFGTRepairHelper.java
Pyritie: ah gotcha

Mqrius: Tbh this isn't really a "data" class so Py does have a point. But also we don't have a good place for core code like this lol

gustovafing: Maybe put it in the utils folder?

Mqrius: It's not really a util either. We need a place for actual core code that's not a mixin, compat, or data or some other category of special code

PR #358, opened 2026-03-13
Fix Blazeburners eating drums

### LINKED ISSUE #3332: [🐛]: Blaze burner consume Gregtech drum alongside with fuel

- [x] No, it was not.

### Computer OS

Windows 11

### Modpack Version Used

0.11.25

### Minecraft Environment

Singleplayer

### New Worlds

Yes

https://mclo.gs/1NlncGt

### Expected Behavior

Burner have fuel inside, empty drum in inventory

### Actual Behavior

Burner have fuel inside, drum consumed

### Steps to Reproduce

1. Take Gregtech drum with liquid fuel (steel with creosote, for example)
2. Right-click to blaze burner
3. Drum consumed alongside with fuel

### Additional Information

TFC barrels don't have this bug

Mqrius:
## What is the new behavior?
Fixes https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3332

## Bug Details
Problem was the original BB code not expecting liquids, so it would just reduce the stack in your hand by one.

## Fix Details
Need to replace most of the liquid fuel insertion handling to use proper IFluidHandler logic instead.

Need to do that fairly high up, where we still have a reference to the original itemstack in hand that we're replacing.

Needs some special handling to deal with fluid-holding items that disappear when empty (buckets) vs ones that don't (GT drums)

## Further Details
Turned off Create Liquid Fuel's own handler since it's now superfluous.

Also remove our own dimension-check mixin since that's also superseded by the new mixin.

You can still insert liquid from barrels into BB on the moon because TFC does its own interaction handling, but it doesn't burn anyways so it's not a problem.

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/mixins/common/create/BlazeBurnerBlockEntityMixin.java
+++ b/src/main/java/su/terrafirmagreg/core/mixins/common/create/BlazeBurnerBlockEntityMixin.java
@@ -16,24 +16,22 @@ import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.entity.BlockEntityType;
 import net.minecraft.world.level.block.state.BlockState;
 
-/**
- * Mixin to prevent fuel from being inserted in the Blaze Burner by hand or mechanical arm in airless dimensions.
- * Needs higher priority than CreateLiquidFuel's mixin to prevent liquid insertion with a bucket.
- * Automated liquid insertion with pipes etc still works, but won't burn.
- */
-@Mixin(value = BlazeBurnerBlockEntity.class, priority = 900, remap = false)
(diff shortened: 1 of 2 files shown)


Pyritie:
> Turned off Create Liquid Fuel's own handler since it's now superfluous.

At that point, do we even still need the original mod?

Mqrius:
It still registers the fluid tank, handles burn time, and does the nbt data handling. We could in-house it if we want, it's not a big mod, but we'd be copying a lot.

Pyritie:
alright fair enough! I didn't know what else it was still doing that we
haven't overridden lol

Pyritie (approved):

PR #364, opened 2026-03-14
TFC 1.21 worldgen

Pyritie:
Don't merge this yet, but here's a PR if anyone wants to give it a look-over


src/main/java/su/terrafirmagreg/core/world/new_ow_wg/biome/TFGBiomes.java
```diff
@@ -0,0 +1,849 @@
...
+                            .spawnable().salty().noSandyRiverShores()));
+    // Sharp, small hills, with lots of water / snaking winding rivers.
+    public static final BiomeExtension LOW_CANYONS = register("low_canyons",
+            riverType(TFGRiverBlendType.WIDE,
+                    builder().heightmap(seed -> BiomeNoise.canyons(seed, -8, 21))
+                            .surface(TFGNormalSurfaceBuilder.INSTANCE)
+                            .aquiferHeightOffset(-16)
+                            .spawnable().noSandyRiverShores()));
+
+    // Mid biomes
```
Redeix, line 107: I think they are cool :(

Xtrial-01:
> if anyone wants to give it a look-over

Who even will be looking through 74 347 lines of code 🥀

Pyritie:
you can skip the jsons lol

Redeix:
> you can skip the jsons lol

you say that like there arent a ton of mixins and classes lol. I gave it a quick look over and nothing really stood out. I think this one will need to be tested live lmao

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/region/TFGChooseBiomesTask.java
```diff
@@ -0,0 +1,302 @@
...
+            } else if (point.baseOceanDepth >= 5 || point.distanceToEdge < 2) {
+                point.biome = DEEP_OCEAN;
+            } else {
+                point.biome = randomSeededFrom(rngSeed, areaSeed, MID_DEPTH_OCEAN_BIOMES);
+            }
+
+            // Add hot spot biomes
+            final byte age = pt.tfg$getHotSpotAge();
+            if (age > 0) {
+                if (age == 4 && point.biome == OCEAN || point.biome == DEEP_OCEAN || point.biome == OCEAN_REEF || point.biome == DEEP_OCEAN_TRENCH) {
```
Mqrius: Operator precedence issue; && binds tighter than ||, so the various biomes should be in brackets

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/layers/TFGMoreShoresLayer.java
```diff
@@ -0,0 +1,48 @@
...
+    public int apply(AreaContext context, int north, int east, int south, int west, int center) {
+        if (center != TFGLayers.OCEAN) {
+            Predicate<IntPredicate> matcher = p -> p.test(north) || p.test(east) || p.test(south) || p.test(west);
+            if (matcher.test(layer -> layer == TFGLayers.TERRACE_LOWER)) {
+                return TFGLayers.TERRACE_UPPER;
+            }
+            if (matcher.test(layer -> layer == TFGLayers.SEA_STACKS)) {
+                return TFGLayers.SEA_STACKS;
+            }
+            if (matcher.test(layer -> layer == TFGLayers.TIDAL_FLATS && layer == TFGLayers.SHORE)) {
```
Mqrius: && should be || I assume, right now it can't be true

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/RockSettingsHelpers.java
```diff
@@ -0,0 +1,35 @@
...
+    static {
+        var presets = RockSettingsAccessor.getPresets();
+
+        isKarstMap.put(presets.get(Helpers.identifier("limestone")), true);
+        isKarstMap.put(presets.get(Helpers.identifier("dolomite")), true);
+        isKarstMap.put(presets.get(Helpers.identifier("chalk")), true);
+        isKarstMap.put(presets.get(Helpers.identifier("marble")), true);
+
+        isMaficMap.put(presets.get(Helpers.identifier("gabbro")), false);
+        isMaficMap.put(presets.get(Helpers.identifier("basalt")), false);
```
Mqrius: I think these two should be `true` instead of `false` since they're both mafic

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/region/TFGChooseRocks.java
```diff
@@ -0,0 +1,59 @@
...
+            var pt = (IRegionPoint) point;
+            // Lower two bits are the supertype, upper bits are seed
+            point.rock = (rockArea.get(pt.tfg$getX(), pt.tfg$getZ()) << TYPE_BITS) | findClosestType(region, point, pt.tfg$getIndex());
+        }
+    }
+
+    private int findClosestType(Region region, Region.Point center, int index) {
+        int type = center.land() ? LAND : OCEAN, minDist = Integer.MAX_VALUE;
+        for (int dx = -2; dx <= 2; dx++) {
+            for (int dz = 0; dz <= 2; dz++) {
```
Mqrius: dz starts at 0, should start at -2

src/main/java/su/terrafirmagreg/core/world/feature/CreepingOceanPlantFeature.java
```diff
@@ -0,0 +1,62 @@
...
+        final BlockPos pos = context.origin();
+        final BlockState state = context.config().block().defaultBlockState();
+        final int radius = context.config().radius();
+        final int height = context.config().height();
+        final BlockPos.MutableBlockPos cursor = new BlockPos.MutableBlockPos();
+
+        for (int x = -radius; x <= radius; x++) {
+            for (int z = -radius; z <= radius; z++) {
+                for (int y = 0; y < height; y++) {
+                    if (x * x + z + z < radius * radius && context.random().nextFloat() < context.config().integrity()) {
```
Mqrius: z + z should be z * z

src/main/java/su/terrafirmagreg/core/world/feature/CreepingOceanPlantFeature.java
```diff
@@ -0,0 +1,62 @@
...
+                                    setBlock(level, cursor, waterloggedState.setValue(TFGBlockProperties.OPEN, isOpen));
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        return false;
```
Mqrius: This method always returns false. Presumably it should return true if the placement succeeded

src/main/java/su/terrafirmagreg/core/world/feature/RotatableWaterPlantFeature.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    public boolean place(FeaturePlaceContext<BlockConfig<RotatableWaterPlantBlock>> context) {
+        final RandomSource random = context.random();
+        final WorldGenLevel level = context.level();
+        final BlockPos pos = context.origin().offset(0, random.nextInt(12), 0);
+        if (!EnvironmentHelpers.isWorldgenReplaceable(level, pos))
+            return false;
+        BlockState state = context.config().block().defaultBlockState();
+        for (Direction direction : Direction.allShuffled(RandomSource.create())) {
```
Mqrius: This uses a new RandomSource which makes it non-deterministic vs world seed. Should probably use existing `random` which ig comes from the seed.

src/main/java/su/terrafirmagreg/core/world/feature/CreepingOceanPlantFeature.java
```diff
@@ -0,0 +1,62 @@
...
+        final int radius = context.config().radius();
+        final int height = context.config().height();
+        final BlockPos.MutableBlockPos cursor = new BlockPos.MutableBlockPos();
+
+        for (int x = -radius; x <= radius; x++) {
+            for (int z = -radius; z <= radius; z++) {
+                for (int y = 0; y < height; y++) {
+                    if (x * x + z + z < radius * radius && context.random().nextFloat() < context.config().integrity()) {
+                        cursor.setWithOffset(pos, x, y, z);
+                        int heightAboveTide = context.config().block().toString().contains("anemone") ? -1 : 2;
```
Mqrius: Do we have a tag or block match for this, instead of string match?

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/WorldgenVersionData.java
```diff
@@ -0,0 +1,102 @@
...
+ */
+public class WorldgenVersionData extends SavedData {
+
+    public static final String DATA_ID = "tfg_worldgen_versions";
+
+    /** Version number for the 1.21-backport overworld worldgen. */
+    public static final int OVERWORLD_TFC_1_21_BACKPORT = 1;
+
+    /** Effective overworld worldgen version for the current server session. Set on server start. */
+    public static int OVERWORLD_VERSION = 0;
```
Mqrius: Okay this one's on me lol, this should be volatile since it's set on the main thread but read by the worldgen threads

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/Seed.java
```diff
@@ -0,0 +1,78 @@
...
+ * from it. Use the original {@code seed} when stability across multiple different noises is needed,
+ * and {@link #next()} otherwise.
+ * <p>
+ * Note in some situations, the only supported access will be via {@link #seed()}, as {@link #next()}
+ * would be unstable, for instance whenever accessed during world generation (after all structures
+ * are initially built, in a predictable sequence).
+ */
+public final class Seed {
+    /** World seed for the current server session, set before RegionGenerator construction. */
+    public static long worldSeed;
```
Mqrius: This one would also be better as volatile, same reason

src/main/java/su/terrafirmagreg/core/common/data/TFGPlant.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    public Block create() {
+        return type.factory.apply(this, type);
+    }
+
+    @Override
+    public int stageFor(Month month) {
+        assert stagesByMonth != null;
+        return stagesByMonth.length < month.ordinal() ? 0 : stagesByMonth[month.ordinal()];
```
Mqrius, line 108: < should be <=, more of a latent bug because all arrays are 12 long anyway.

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/noise/TFGBiomeNoise.java
```diff
@@ -0,0 +1,1035 @@
...
+
+    /**
+     * Effectively a {@code lerp(noiseA(), noiseB(), piecewise(noiseB()) + noiseC()} with the following additional techniques:
+     * <ul>
+     *     <li>{@code noiseA} is scaled to outside it's range, then biased towards 1.0, to expose more cliffs, as opposed to hills </li>
+     *     <li>{@code piecewise()} is a piecewise linear function that creates cliff shapes from the standard noise distribution.</li>
+     *     <li>{@code noiseC} is added on top to provide additional variance (in places where the piecewise function would otherwise flatten areas.</li>
+     * </ul>
+     */
+    public static Noise2D sharpHills(long seed, float minMeight, float maxHeight) {
```
Mqrius: minMeight > minHeight

src/main/java/su/terrafirmagreg/core/mixins/common/tfc/new_ow_wg/BiomeBuilderMixin.java
```diff
@@ -0,0 +1,142 @@
...
+
+    public BiomeBuilder tfg$tuffRings(int frequency, int baseHeight, int scaleHeight) {
+        this.tfg$hasTuffRings = true;
+        this.volcanoFrequency = frequency;
+        this.tfg$centeredFeatureBaseHeight = baseHeight;
+        this.tfg$centeredFeatureScaleHeight = scaleHeight;
+        return (BiomeBuilder) (Object) this;
+    }
+
+    public BiomeBuilder tfg$tuyas(int frequency, int baseHeight, int scaleHeight, int tuyaBasaltHeight, boolean icy) {
```
Mqrius, line 117: boolean icy is passed to this method, but `this.tfg$centeredFeatureIce` is not set. So CenteredFeatureBuilder never runs `calculateIcyShape` instead of calculateShape

src/main/java/su/terrafirmagreg/core/mixins/common/tfc/new_ow_wg/BiomeBuilderMixin.java
```diff
@@ -0,0 +1,142 @@
...
+        this.tfg$hasTuyas = true;
+        this.volcanoFrequency = frequency;
+        this.tfg$centeredFeatureRockHeight = SEA_LEVEL_Y + tuyaBasaltHeight;
+        this.tfg$centeredFeatureBaseHeight = baseHeight;
+        this.tfg$centeredFeatureScaleHeight = scaleHeight;
+        return (BiomeBuilder) (Object) this;
+    }
+
+    public BiomeBuilder tfg$cinderCones(int frequency, int baseHeight, int scaleHeight, int cinderConeBasaltHeight, boolean additive) {
```
Mqrius, line 127: boolean additive is unused, not sure what its purpose was supposed to be

src/main/java/su/terrafirmagreg/core/mixins/common/tfc/new_ow_wg/BiomeBuilderMixin.java
```diff
@@ -0,0 +1,142 @@
...
+    @Inject(method = "surface", at = @At("TAIL"), remap = false)
+    private void tfg$surface(SurfaceBuilderFactory surfaceBuilderFactory, CallbackInfoReturnable<BiomeBuilder> cir) {
+        if (OVERWORLD_VERSION == OVERWORLD_TFC_1_21_BACKPORT) {
+            this.surfaceBuilderFactory = CinderConeSurfaceBuilder.create(surfaceBuilderFactory);
+            this.surfaceBuilderFactory = TuffRingsSurfaceBuilder.create(this.surfaceBuilderFactory);
+            this.surfaceBuilderFactory = TuyaSurfaceBuilder.create(this.surfaceBuilderFactory);
+        }
+    }
+
+    public BiomeBuilder tfg$tuffRings(int frequency, int baseHeight, int scaleHeight) {
```
Mqrius, line 109: tfg$tuffRings doesn't set this.volcanic = true. Not sure if intentional?

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/surface_builders/TuyaSurfaceBuilder.java
```diff
@@ -0,0 +1,79 @@
...
+
+        int surfaceDepth = -1;
+        for (int y = startY; y >= endY; --y) {
+            BlockState stateAt = context.getBlockState(y);
+            if (stateAt.isAir()) {
+                // Reached air, reset surface depth
+                surfaceDepth = -1;
+            } else if (context.isDefaultBlock(stateAt)) {
+                IBiomeExtension be = (IBiomeExtension) context.biome();
+                if (y > be.tfg$getCenteredFeatureRockHeight() + noise) {
```
Mqrius, line 66: This should probably be `ctx.tfg$getTuyaBiome()` instead of context.biome

src/main/java/su/terrafirmagreg/core/mixins/common/tfc/new_ow_wg/BiomeBuilderMixin.java
Mqrius: This breaks feature placement but not rock/tuff placement iiuc, so if you're missing trees...

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/rivers/TFGRiverBlendType.java
```diff
@@ -0,0 +1,38 @@
...
+package su.terrafirmagreg.core.world.new_ow_wg.rivers;
+
+import java.util.function.Function;
+
+import su.terrafirmagreg.core.world.new_ow_wg.Seed;
+
+public enum TFGRiverBlendType {
+    NONE(seed -> TFGRiverNoiseSampler.NONE),
+    BANKED(TFGRiverNoise::banked), // Raised banks 1-2 blocks above water that may be higher than surrounding terrain in swampy biomes
+    TALL_BANKED(TFGRiverNoise::tallBanked), // Sim to banked, but taller for use in mud flat/salt flat biomes
```
Mqrius, line 16: TALL_BANKED is never used, probably a TFC bug

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/region/TFGChooseBiomesTask.java
Pyritie: issue in tfc, but that does look like a legit bug

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/layers/TFGMoreShoresLayer.java
Pyritie: same

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/RockSettingsHelpers.java
Pyritie: fixed

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/region/TFGChooseRocks.java
Pyritie: it's like that in base tfc

src/main/java/su/terrafirmagreg/core/world/feature/CreepingOceanPlantFeature.java
Pyritie: base tfc

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/chunk/TFGChunkNoiseFiller.java
```diff
@@ -0,0 +1,612 @@
...
+            }
+        }
+    }
+
+    /**
+     * @param y The y position
+     * @param heightNoiseValue The calculated average height noise, from {@link BiomeNoiseSampler#height()}
+     * @return The density noise for the given y position, where positive noise is solid, in the range [0, 1]
+     */
+    private double calculateNoiseAtHeight(int y, double heightNoiseValue) {
```
Mqrius, line 423: There's something off here. The river and the shore both add to the noise without any reset inbetween, but that results in the noise summing up to 2 instead of to 1. It's notably different from how TFGChunkHeightFiller does it.  The practical result would be that it's easier for AIR_THRESHOLD to be exceeded (it's as if it's set to 0.2 instead of 0.4), which would make more blocks air. It's also harder for cave carvers to cross the threshold, so caves are rarer and smaller than intended.

src/main/java/su/terrafirmagreg/core/world/feature/CreepingOceanPlantFeature.java
Pyritie: base tfc, but made the change

src/main/java/su/terrafirmagreg/core/world/feature/RotatableWaterPlantFeature.java
Pyritie: base tfc

src/main/java/su/terrafirmagreg/core/world/feature/CreepingOceanPlantFeature.java
Pyritie: no tag, though I could add one. There's four anemone blocks. Would using a tag be better?

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/WorldgenVersionData.java
Pyritie: done

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/Seed.java
Pyritie: this is your addition right? the tfc class doesn't have it

src/main/java/su/terrafirmagreg/core/world/feature/CreepingOceanPlantFeature.java
Mqrius: It works this way but it's kinda hacky. Tags or direct comparisons are the proper way to check the identity of a block. But I'm not sure whether they're available at this point so maybe it doesn't work?

src/main/java/su/terrafirmagreg/core/common/data/TFGPlant.java
Pyritie: that's the same as tfc 1.20 code this time lol

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/noise/TFGBiomeNoise.java
Pyritie: tfc, fixed

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/Seed.java
Mqrius: Yeah, just to get all the worldSeed access from a single source of truth. I just didn't remember that it's accessed from multiple threads. If you make it volatile you can guarantee that all threads will see the same value and won't get stale reads.

src/main/java/su/terrafirmagreg/core/mixins/common/tfc/new_ow_wg/BiomeBuilderMixin.java
Pyritie: my bad, must've missed it when copying

src/main/java/su/terrafirmagreg/core/mixins/common/tfc/new_ow_wg/BiomeBuilderMixin.java
Pyritie: it's unused in tfc too. Maybe it's used for some future thing

src/main/java/su/terrafirmagreg/core/mixins/common/tfc/new_ow_wg/BiomeBuilderMixin.java
Pyritie: Intentional -- 1.21 renamed the `volcanic` field to `hasCinderCones`, and only the `cinderCones()` method sets that to true, not the tuff rings or tuyas.

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/chunk/TFGChunkNoiseFiller.java
Mqrius: But maybe all the other values are already jiggled such that this doesn't show up as a bug. If the worldgen looks good this way then you could leave it.

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/surface_builders/TuyaSurfaceBuilder.java
Pyritie: tfc just calls `context.biome().getCenteredFeatureRockHeight()`, so possibly intentional?

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/layers/TFGMoreShoresLayer.java
```diff
@@ -0,0 +1,48 @@
...
+import net.dries007.tfc.world.layer.framework.AreaContext;
+
+import su.terrafirmagreg.core.world.new_ow_wg.TFGLayers;
+
+public enum TFGMoreShoresLayer implements AdjacentTransformLayer {
+    INSTANCE;
+
+    @Override
+    public int apply(AreaContext context, int north, int east, int south, int west, int center) {
+        if (center != TFGLayers.OCEAN) {
```
Mqrius, line 22: this should be `!TFGLayers.isOcean(center)` because now ocean reef and other non-standard oceans can get converted to shore.

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/rivers/TFGRiverBlendType.java
Pyritie: looks like it's used for mud flats and salt flats in 1.21, fixed. These were the first two biomes I ported over before touching river code, so I guess I forgot to change them to the new river types once that was done.

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/chunk/TFGChunkNoiseFiller.java
Pyritie: It looks the same as what tfc does, so I don't know how intentional it is.

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/chunk/TFGChunkHeightFiller.java
```diff
@@ -0,0 +1,370 @@
...
+    /**
+     * For a given (x, z) position, samples the provided biome weight map to calculate the height at that location, and the biome
+     *
+     * @param useCache If, in the stateful implementation, arrays corresponding to position within the chunk should be updated.
+     * @return The maximum height at this location
+     */
+    protected final double sampleColumnHeightAndBiome(Object2DoubleMap<BiomeExtension> biomeWeights, boolean useCache) {
+        columnBiomeNoiseSamplers.clear();
+
+        double height = 0, normalHeight = 0, shoreHeight = 0, shoreBaseHeight = 0, oceanHeight = 0;
```
Mqrius: shoreBaseHeight and oceanHeight are written to but never read

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/layers/TFGMoreShoresLayer.java
Pyritie: Unsure, because the two deep ocean biomes shouldn't be near shores in the first place. I don't know about reefs.

src/main/java/su/terrafirmagreg/core/world/feature/CreepingOceanPlantFeature.java
Pyritie: They are, features happen pretty late in the worldgen process, way after things like blocks/items/tags/recipes

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/surface_builders/TuyaSurfaceBuilder.java
Mqrius: I think that might get you the incorrect height though. Not sure, lemme look closer

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/chunk/TFGChunkHeightFiller.java
Pyritie: same as tfc

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/surface_builders/TuyaSurfaceBuilder.java
Mqrius: aiui:

So basically right now in the central biome where the tuya lives, getCenteredFeatureRockHeight works fine, placing basalt down to sea level + 6.

But the tuya extends into neighboring biomes, and there getCenteredFeatureRockHeight will return a default of zero. That makes this place basalt down to bedrock in neighboring biomes.

Pyritie: we can try it!

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/layers/TFGMoreShoresLayer.java
Mqrius: Mm yeah okay true, seems like it should always be oceans next to shores, simply from the baseOceanDepth being low. Though the extra guard wouldn't hurt.
Reefs are from mid ocean, baseOceanDepth 3 to 4, so probably also safe

src/main/java/su/terrafirmagreg/core/world/feature/CreepingOceanPlantFeature.java
Mqrius: Seems a bug though, it's trying to define a circle but right now it isn't.

src/main/java/su/terrafirmagreg/core/world/new_ow_wg/region/TFGChooseRocks.java
Mqrius: Also a bug, rn it doesn't look north to find the closest type

Pyritie: ok

src/main/java/su/terrafirmagreg/core/world/feature/CreepingOceanPlantFeature.java
Pyritie: ok

PR #372, opened 2026-03-21
Migrating the repair system to Core-Modern

Thomasx0:
item repair recipes (tfg:item_repair) with KubeJS
Adds a custom crafting recipe type tfg:item_repair: repair damageable items using a fixed 3×3 pattern (roles H / R / T / C), with repair amount controlled by repairPercentage


src/main/java/su/terrafirmagreg/core/common/data/recipes/repair/ItemRepairRecipeSerializer.java
```diff
@@ -0,0 +1,116 @@
...
+import com.google.gson.JsonObject;
+
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.item.crafting.CraftingBookCategory;
+import net.minecraft.world.item.crafting.Ingredient;
+import net.minecraft.world.item.crafting.RecipeSerializer;
+
+/**
+ * Loads and saves tfg:item_repair. JSON must have pattern and key; repairPercentage is optional and defaults to 0.75.
```
Pyritie: would prefer if the default is changed to 0.25 since the repair kits are cheap and that's what we use for other tools

src/main/java/su/terrafirmagreg/core/common/data/recipes/repair/RepairRecipeHelper.java
```diff
@@ -0,0 +1,117 @@
...
+import net.minecraftforge.registries.ForgeRegistries;
+
+/** Small helpers for item repair: tool tags, repair ingredients, metal ingots, applying durability loss. */
+public class RepairRecipeHelper {
+
+    private static final List<TagKey<Item>> TOOL_ITEM_TAGS = buildToolItemTags();
+
+    private static List<TagKey<Item>> buildToolItemTags() {
+        String[] ids = {
+                "tfc:hammers",
```
Pyritie: don't need two hammer tags

src/main/java/su/terrafirmagreg/core/common/data/recipes/repair/RepairRecipeHelper.java
```diff
@@ -0,0 +1,117 @@
...
+            return true;
+        }
+        if (metalType != null && isMetalIngot(stack, metalType)) {
+            return true;
+        }
+        return false;
+    }
+
+    /** True if the stack is in the forge ingot tag for that metal or is the TFC metal ingot item for it. */
+    public static boolean isMetalIngot(ItemStack stack, @Nullable TFGMetalType metalType) {
```
Pyritie: what is this code for?
we're mostly going to be using repair kits as the consumed item, which you can get via `#forge:repairKit/x` or `ChemicalHelper.get(TFGTagPrefix.repairKit, material, 1)`

src/main/java/su/terrafirmagreg/core/common/data/recipes/repair/TFGMetalType.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.Locale;
+
+/**
+ * Metal ids used with forge ingot tags, TFC ingot item ids, and similar paths. Each constant maps to a lowercase name via
+ * getSerializedName(), same style as datapacks and scripts. Actual repair recipes are defined elsewhere; this enum is only a
+ * shared list of materials.
+ */
+
+public enum TFGMetalType {
```
Pyritie: we have some materials that are only in kubejs that we might want repair recipes for, as well as some things like clothes and space suits that will want unrelated items like cloth. Can the repair system handle this?

src/main/java/su/terrafirmagreg/core/compat/emi/ItemRepairEmiRecipe.java
```diff
@@ -0,0 +1,124 @@
...
+        return 116;
+    }
+
+    @Override
+    public int getDisplayHeight() {
+        return 54;
+    }
+
+    @Override
+    public void addWidgets(@NotNull dev.emi.emi.api.widget.WidgetHolder holder) {
```
Pyritie: why not just import this instead of the full namespace?

Pyritie (changes requested):

src/main/java/su/terrafirmagreg/core/common/data/recipes/repair/ItemRepairRecipeSerializer.java
Thomasx0: good point, updated

src/main/java/su/terrafirmagreg/core/common/data/recipes/repair/RepairRecipeHelper.java
Thomasx0: Agreed, cleaned

src/main/java/su/terrafirmagreg/core/common/data/recipes/repair/RepairRecipeHelper.java
Thomasx0: You’re right this was extra complexity for the old metal-ingot fallback path.
I removed the TFGMetalType-based fallback and kept the recipe system Ingredient-driven, so repair inputs are fully defined by recipe key.R
I think it will be better this way

Pyritie: awesome, thanks

src/main/java/su/terrafirmagreg/core/common/data/recipes/repair/TFGMetalType.java
Thomasx0: It works for more than just metals.

I added a test tfg:item_repair recipe for ad_astra:space_suit, using tfg:space_suit_fabric as the repair material, and the suit repairs correctly.

The existing core repair-kit flow also still works and now restores 25% durability.

The only thing that is still not perfect is the EMI preview: in some cases it may show a durability value different from the final crafting result. This is a preview/display limitation, not a repair logic issue. I probably won’t be able to fix that one (java noob xD).
Actual crafting behavior is correct, and repair kits still clearly state that they restore 25% durability.
изображение
изображение
изображение
изображение

Pyritie: ok perfect, that sounds fine then

src/main/java/su/terrafirmagreg/core/common/data/recipes/repair/ItemRepairRecipeSerializer.java
```diff
@@ -27,7 +27,7 @@ public ItemRepairRecipe fromJson(ResourceLocation recipeId, JsonObject json) {
                 json.has("category") ? json.get("category").getAsString() : "misc",
                 CraftingBookCategory.MISC);
 
-        float repairPercentage = json.has("repairPercentage") ? json.get("repairPercentage").getAsFloat() : 0.75f;
+        float repairPercentage = json.has("repairPercentage") ? json.get("repairPercentage").getAsFloat() : 0.25f;
```
Pyritie: could use DEFAULT_REPAIR_PERCENT here too

Thomasx0: done

Pyritie (approved):

gustovafing (approved):

PR #375, opened 2026-03-22
tfc:drinkable Tooltip

SakuraKitsurugi:
### What is the new behavior?
Drinkables that have effects and have `#tfc:drinkables` tag now have effect tooltip. Excludes liquids with `tfcagedalcohols` namespace.

### Implementation Details
- ForgeClientListened.java

### Additional Information
- I used `.is()` which, apparently, is deprecated (`.builtInRegistryHolder()` also, but did not use it here).
- @gustovafing
- If the PR will be merged I will also need to fix some tags through KubeJS.

### Potential Compatibility Issues
- Technically my first "actual" Java code. I would appreciate info on what could be improved.

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/client/ForgeClientEventListener.java
+++ b/src/main/java/su/terrafirmagreg/core/client/ForgeClientEventListener.java
@@ -14,11 +14,11 @@ import net.minecraft.resources.ResourceLocation;
 import net.minecraft.tags.TagKey;
 import net.minecraft.util.Mth;
 import net.minecraft.util.StringUtil;
-import net.minecraft.world.effect.MobEffect;
 import net.minecraft.world.effect.MobEffectInstance;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.level.ChunkPos;
 import net.minecraft.world.level.Level;
+import net.minecraft.world.level.material.Fluid;
 import net.minecraftforge.api.distmarker.Dist;
 import net.minecraftforge.api.distmarker.OnlyIn;
 import net.minecraftforge.common.capabilities.ForgeCapabilities;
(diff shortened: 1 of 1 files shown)


src/main/java/su/terrafirmagreg/core/client/ForgeClientEventListener.java
```diff
@@ -118,11 +126,27 @@ public static void onItemTooltip(ItemTooltipEvent event) {
...
+                        drink.getEffects().forEach(effect -> event.getToolTip().add(getTooltip(effect.type(), effect.duration(), effect.amplifier())));
+                    }
+                }
+            });
         }
     }
 
     // These are taken from TFC Aged Alcohol
 
+    private static Component getTooltip(MobEffect effect, int duration, int amplifier) {
```
Pyritie: you can reuse the method right below this

src/main/java/su/terrafirmagreg/core/client/ForgeClientEventListener.java
```diff
@@ -118,11 +126,27 @@ public static void onItemTooltip(ItemTooltipEvent event) {
             if (foodProperties != null) {
                 foodProperties.getEffects().forEach(effect -> event.getToolTip().add(getTooltip(effect.getFirst())));
             }
+
+            var alcoholTag = TagKey.create(Registries.FLUID, ResourceLocation.fromNamespaceAndPath("tfc", "drinkables"));
```
Pyritie: this tag should be cached into a private field or something, because otherwise this method is called every time a new item is hovered over

Pyritie (changes requested):

src/main/java/su/terrafirmagreg/core/client/ForgeClientEventListener.java
SakuraKitsurugi: Sadly, I can't. Liquids/Fluids have different effect properties for some reason.

Pyritie: can't you pass the new MobEffectInstance into it?

SakuraKitsurugi: Ah, right. Lemme check

Pyritie (approved):

gustovafing (approved):

PR #379, opened 2026-03-23
Fluid vein fix

TomPlop:


.idea/misc.xml
TomPlop, line 1: what is that?

src/main/resources/data/tfg/tags/block/dry_plant_plantable_on.json
TomPlop, line 1: bad merge?

src/main/resources/data/tfg/tags/block/epiphyte_plantable_on.json
TomPlop, line 1: bad merge?

src/main/resources/data/tfg/tags/block/sea_stack_rocks.json
TomPlop, line 1: bad merge?

.idea/misc.xml
Pyritie, line 1: delete this one

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/trait/GasWellRecipeLogic.java
```diff
@@ -0,0 +1,167 @@
...
+            return;
+
+        // Check the vein
+        var veinFluid = entry.getDefinition().getStoredFluid().get();
+        if (veinFluid == null)
+            return;
+
+        // Only work for natural_gas
+        var naturalGas = net.minecraftforge.registries.ForgeRegistries.FLUIDS
+                .getValue(net.minecraft.resources.ResourceLocation.fromNamespaceAndPath("gtceu", "natural_gas"));
```
Pyritie: There's a `GTCEu.id` I think? that you can use instead

Pyritie: but also I think you can just query the natural gas material itself for its fluid

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/trait/GasWellRecipeLogic.java
```diff
@@ -0,0 +1,167 @@
...
+            if (!stack.isEmpty() && isExplosive(stack)) {
+                itemHandler.extractItemInternal(i, 1, false);
+                hasConsumedExplosive = true;
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private boolean isExplosive(net.minecraft.world.item.ItemStack stack) {
```
Pyritie: this can be an import at the top

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/trait/GasWellRecipeLogic.java
```diff
@@ -0,0 +1,167 @@
...
+        if (stack.is(Items.TNT))
+            return true;
+        var dynamite = net.minecraftforge.registries.ForgeRegistries.ITEMS
+                .getValue(net.minecraft.resources.ResourceLocation.fromNamespaceAndPath("gtceu", "powderbarrel"));
+        if (dynamite != null && stack.is(dynamite))
+            return true;
+        return stack.is(net.minecraft.tags.ItemTags.create(
```
Pyritie: this tag should be in our TFGTags file

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/trait/GasWellRecipeLogic.java
```diff
@@ -0,0 +1,167 @@
...
+            }
+        }
+        return false;
+    }
+
+    private boolean isExplosive(net.minecraft.world.item.ItemStack stack) {
+        if (stack.is(Items.TNT))
+            return true;
+        var dynamite = net.minecraftforge.registries.ForgeRegistries.ITEMS
+                .getValue(net.minecraft.resources.ResourceLocation.fromNamespaceAndPath("gtceu", "powderbarrel"));
```
Pyritie: `GTCEu.id`

Pyritie: also why not just get the item from gt directly? GTItems or whatever

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
```diff
@@ -27,10 +39,26 @@
...
 
+    private static final Logger LOGGER = LogManager.getLogger();
+
+    private static Block cachedDryIce = null;
+    private static Block cachedGeyserite = null;
+
+    private static boolean loadGasVentBlocks() {
+        if (cachedDryIce != null && cachedGeyserite != null)
+            return true;
+        cachedDryIce = ForgeRegistries.BLOCKS
```
Pyritie: you can just use the TFGBlocks classes directly...

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
```diff
@@ -65,6 +94,178 @@ public static void onLevelUnload(LevelEvent.Unload event) {
...
+                    .registry(Registries.CONFIGURED_FEATURE)
+                    .orElse(null);
+            if (cfRegistry == null)
+                return;
+
+            var configuredFeature = cfRegistry.get(featureId);
+            if (configuredFeature == null)
+                return;
+
+            if (!(configuredFeature.feature() instanceof com.gregtechceu.gtceu.common.worldgen.feature.FluidSproutFeature))
```
Pyritie: import these instead of writing the whole namespace

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
```diff
@@ -65,6 +94,178 @@ public static void onLevelUnload(LevelEvent.Unload event) {
         }
     }
 
+    @SubscribeEvent
+    public static void onAddReloadListeners(AddReloadListenerEvent event) {
+        event.addListener(BedrockFluidSpoutLoader.INSTANCE);
+    }
+
+    @SubscribeEvent
+    public static void onChunkLoad(ChunkEvent.Load event) {
```
Pyritie, line 82: does this happen every time the chunk is loaded? or just when it's first generated? because otherwise you're going to be recreating this feature every time, overwriting anything that players might have changed

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
```diff
@@ -65,6 +94,178 @@ public static void onLevelUnload(LevelEvent.Unload event) {
...
+            var configuredFeature = cfRegistry.get(featureId);
+            if (configuredFeature == null)
+                return;
+
+            if (!(configuredFeature.feature() instanceof com.gregtechceu.gtceu.common.worldgen.feature.FluidSproutFeature))
+                return;
+
+            var config = (com.gregtechceu.gtceu.common.worldgen.feature.configurations.FluidSproutConfiguration) configuredFeature.config();
+
+            serverLevel.getServer().execute(() -> {
```
Pyritie: I don't like this whole approach. It shouldn't be setting blocks as you load chunks.
Instead, this should be a feature that's placed in every chunk, and then the feature itself decides whether it has the correct fluid or not.
This is not only a lot cleaner, but also only happens once when the feature is created, and not every time the chunk is loaded as the player walks around.

src/main/java/su/terrafirmagreg/core/world/BedrockFluidSpoutLoader.java
```diff
@@ -0,0 +1,82 @@
...
+    protected void apply(Map<ResourceLocation, JsonElement> objects,
+            ResourceManager manager,
+            ProfilerFiller profiler) {
+
+        VEIN_TO_FEATURE.clear();
+        VEIN_TO_TYPE.clear();
+
+        for (var fileEntry : objects.entrySet()) {
+            if (!fileEntry.getValue().isJsonObject()) {
+                LOGGER.warn("[TFG] bedrock_fluid_spouts: fichier {} n'est pas un objet JSON, ignoré",
```
Pyritie: english

src/main/resources/data/tfg/tags/block/dry_plant_plantable_on.json
Pyritie: yeah these tag files shouldn't exist, they were moved to `blocks` instead of `block`

Pyritie (changes requested):

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
TomPlop: It doesn't it only run the event fully if it's a new chunk

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/trait/GasWellRecipeLogic.java
TomPlop: didn't allow me because I used getStoredFluid and IDEA was giving me an error if I used a material.

I can try with GTCEu.id

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/trait/GasWellRecipeLogic.java
TomPlop: GTItems doesn't work for the powder barrel I tried GTBlocks too never found it.

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/trait/GasWellRecipeLogic.java
TomPlop: Much cleaner now

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/trait/GasWellRecipeLogic.java
TomPlop: It works with GTceu.id

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/trait/GasWellRecipeLogic.java
Pyritie: why doesn't it work?

TomPlop: I've no idea but I changed it for something much cleaner using a tag that I added in the other PR

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
TomPlop: So I loaded my previous code. I think the main issue is that GTm doesn't show the fluid veins until they are either prospected or a fluid rig is placed. I had to manually saved them when generating a chunk and I think it saves them after the place feature are called from TFC which make TFC looks for a feature that it can't generate because it doesn't know yet which fluid exists in the vein.

I honestly don't know how to change that it seems like there is a conflict between GTm fluid vein and TFC feature. So whenever we try to use the TFC feature system it will never load.

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
TomPlop: I don't find Geyserite

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
Pyritie: Sounds like we'd need to force GT to generate the vein earlier, maybe gustavo can help with that? Or at least have it generate whenever the feature checks for it

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
Pyritie: ok we can move that block into core

src/main/java/su/terrafirmagreg/core/common/ForgeCommonEventListener.java
TomPlop: Yeah that's the only option. Basically right now I do it onChunkLoad which is definitely way too late. The thing is even GTm stores them quite late so either doing it on a different method earlier could work but if GTm does it too late then a mixin to make it start earlier. I honestly have no idea how though.

The other possibility would be to fully copy GTm but do our own.

TomPlop: But that brings a lot of issues by needing to make our own prospector and our own fluid rig I guess

TomPlop:
Everything else was fix that only thing would be to fully redo the Spout/Gas Indicator placement if we can make it work through tfc feature

src/main/java/su/terrafirmagreg/core/common/data/tfgt/machine/trait/GasWellRecipeLogic.java
```diff
@@ -146,15 +152,10 @@ private boolean consumeExplosive() {
...
-    private boolean isExplosive(net.minecraft.world.item.ItemStack stack) {
-        if (stack.is(Items.TNT))
-            return true;
-        var dynamite = net.minecraftforge.registries.ForgeRegistries.ITEMS
-                .getValue(net.minecraft.resources.ResourceLocation.fromNamespaceAndPath("gtceu", "powderbarrel"));
-        if (dynamite != null && stack.is(dynamite))
-            return true;
-        return stack.is(net.minecraft.tags.ItemTags.create(
-                net.minecraft.resources.ResourceLocation.fromNamespaceAndPath("tfg", "explosives")));
+    private static final TagKey<Item> EXPLOSIVES = ItemTags.create(ResourceLocation.fromNamespaceAndPath("tfg", "explosives"));
```
Pyritie, line 155: this should be a new tag in TFGTags

TomPlop: Oh I added into Modern repo but I guess I could add it in core

gustovafing (approved):

src/main/java/su/terrafirmagreg/core/world/feature/FluidGasVentFeature.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public FluidGasVentFeature(Codec<FluidGasVentConfig> codec) {
+        super(codec);
+    }
+
+    private static boolean loadBlocks() {
+        if (cachedDryIce != null && cachedGeyserite != null)
+            return true;
+        cachedDryIce = TFGBlocks.DRY_ICE.get();
+        cachedGeyserite = ForgeRegistries.BLOCKS
```
Pyritie, line 30: we should move this into core

Pyritie (approved):
looks much cleaner separated out like this!

PR #396, opened 2026-03-30
new biome tag and update vein tab

TomPlop:


Pyritie (approved):

src/main/resources/data/tfg/tags/worldgen/biome/earth/is_hill.json
```diff
@@ -0,0 +1,19 @@
+{
+    "values": [
+        "tfg:earth/hills",
```
Pyritie, line 3: what about plateaus?

TomPlop: I was deeply hesitating but then it's a lot like is_normal

Also hills has this cool thing where the veins can spawn on the side and you see them where that won't be true with plateau

Pyritie:
I guess

PR #404, opened 2026-04-04
Cache noise objects instead of recreating every call

Mqrius:
## Details
Cache the noise objects around shore generation. They're expensive to generate anew and they are the same every time, so it's good to cache them.

Shouldn't change anything in terms of behavior or determinism.

## Performance Testing
New Minecraft world, seed '1'
Wait 2 minutes for things to stabilize

`/forge generate 3642 77 1449 10000`
Time with stopwatch

#### Before:
run 1: 5:27
run 2: 5:33

#### With this PR:
run 1: 5:09
run 2: 5:05

So about a 10% improvement in CPU time, possibly also some improvement in memory GC pressure, varying depending on whether you hit a lot of shore chunks

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/world/new_ow_wg/noise/ShoreNoise.java
+++ b/src/main/java/su/terrafirmagreg/core/world/new_ow_wg/noise/ShoreNoise.java
@@ -22,13 +22,14 @@ public final class ShoreNoise {
     // Typical monoslope beach
     public static ShoreNoiseSampler sandyBeach(Seed seed) {
         return new ShoreNoiseSampler() {
+            final Noise2D tideLevelNoise = TFGBiomeNoise.shoreTideLevelNoise(seed);
             double sandHeight, weight;
 
             @Override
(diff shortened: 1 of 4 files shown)


Pyritie (commented):
Why is the lava flow noise moved to the shore builder?

I'm very wary towards any sort of worldgen optimization stuff right now because there's going to be a whole heap of changes in a month or two from tfc's end. And the more things we change, the harder it'll be to figure out how to backport it all

Mqrius:
Lava flow; that wasn't me that was already there:
Screenshot_2026-04-05-19-18-14-64_40deb401b9ffe8e1df2f1cc5ba480b12

As for the upcoming changes, I suppose this could be a PR to tfc in theory. It's a bit weird to me that they didn't already make these noise objects finals since they do do it with a bunch of other ones.

Pyritie:
so it's github being weird again?

Mqrius:
No the lava flows are in the shore builder because the shore builder does something when lava flows are present. I didn't move it there, I just cache the noise objects.

Pyritie:
ooh ok

Pyritie (approved):

PR #410, opened 2026-04-05
Fix beardifier plateaus: Force terrain noise slightly negative above normal surface height.

Mqrius:
## Bug details
Due to rounding and stuff, the terrain noise can actually be slightly above zero in the block above the surface.

Normally this would cause a block to be placed, but TFC doesn't actually iterate that Y value normally so it gets cut off.

With the beardifier needing to iterate additional Y values, this caused a bunch of additional blocks to be placed on the surface just from the terrain noise, if a structure was nearby.

## Fix details
If we're dealing with a Y value that's above the normal surface height, we clamp the terrain noise value to negative 0.05 ( == slightly biased towards air). That way the beardifier can still pull it up to solid terrain if the structure is nearby enough.


Pyritie (approved):

Pyritie:
thank you for helping with this!

PR #415, opened 2026-04-08
Add better fluid vein emi page

### LINKED ISSUE #3605: [💎]: Add new EMI tab for bedrock fluid veins
### Was this enhancement suggested before?

- [x] No, it was not.

### What would you like to be added/ changed?

Add a new EMI tab for bedrock fluid veins

### Why is this needed?

Since our fluid veins are now affected by climate and biome, the GT fluid vein tab is insufficient.

BlueBoat29:
could use color


src/main/java/su/terrafirmagreg/core/common/tfgt/worldgen/ClimateWeightModifier.java
```diff
@@ -35,6 +35,20 @@ public ClimateWeightModifier(Mode mode, float min, float max, int addedWeight) {
         this.addedWeight = addedWeight;
     }
 
+    /// Returns Map keyed by "temperature" and/or "rainfall"
+    /// @return Map of List(float) containing minValue, maxValue
+    public Map<String, List<Float>> getClimates() {
```
Pyritie: any reason why you convert the mode into a string instead of just keeping it as an enum?

src/main/java/su/terrafirmagreg/core/compat/emi/FluidVeinRecipe.java
```diff
@@ -0,0 +1,349 @@
...
+                Component.translatable("tfg.emi.fluid_veins.depletion.tooltip", depChance + "%", depAmount, BedrockFluidVeinSavedData.MAXIMUM_VEIN_OPERATIONS));
+
+        yPointer = addTextLine(widgets, Component.translatable("tfg.emi.fluid_veins.depleted_yield", depYield + "mB/s"), 2, yPointer);
+
+        if (weight != 0) {
+            yPointer = addTextLine(widgets, Component.translatable("tfg.emi.fluid_veins.weight", weight), 2, yPointer);
+        }
+
+        //yPointer = addInlineDimensionIcons(widgets, 2, yPointer);
+        //Still don't know which one looks better
```
Pyritie, line 158: my vote is for the corner!

src/main/java/su/terrafirmagreg/core/common/tfgt/worldgen/ClimateWeightModifier.java
BlueBoat29: enums are scary

PR #419, opened 2026-04-09
port of vacuum chamber recipe speed fix

ashleney:
https://github.com/SSWTLZZ69/Create-Vintage-Improvements/pull/22

I do not know how to remove the cast to an integer but I don't think it should cause any particular issues. I tried my best

see linked PR for info on why this is a useful change

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/mixins/common/vintageimprovements/VacuumChamberBlockEntityMixin.java
+++ b/src/main/java/su/terrafirmagreg/core/mixins/common/vintageimprovements/VacuumChamberBlockEntityMixin.java
@@ -6,6 +6,7 @@ import org.spongepowered.asm.mixin.injection.ModifyVariable;
 
 import com.negodya1.vintageimprovements.content.kinetics.vacuum_chamber.VacuumChamberBlockEntity;
 
+// NOTE: Delete once https://github.com/SSWTLZZ69/Create-Vintage-Improvements/pull/22 gets merged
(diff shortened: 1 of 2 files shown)


Pyritie:
what cast to an integer?

ashleney:
the original code casts to an integer for no reason so it's not a smooth function, I have no idea how to edit that with a mixin

Pyritie:
oh, well it's not a huge deal

Pyritie:
could you write a comment saying when this mixin should be deleted, with a link to your PR? so we don't forget what this does

ashleney:
added

Pyritie (approved):

ashleney:
Before you merge this, I implore you to think hard if it's even worth it trying to keep parity with the formula that vintageimprovements does.

Currently all recipes that have a processing time of >3000 take 25.6s iregardless of RPM.
This change makes it so that running it at <128RPM makes it take 25.6s and at 256RPM decrease to 12.8s. This makes it so that parity is kept with fast VI recipes but still lacks intuitive sense, because 2 vacuum chambers at 1RPM are the same speed as 1 vacuum chamber at 256RPM. 

Maybe removing the time cap could help? All vacuum chamber recipes would have to have their duration decreased.
Maybe duration should just be ticks and there should be no logarithm? It'll work completely different but I also think it would be the best choice for the future.

Pyritie:
me? think hard?

Redeix (approved):

PR #430, opened 2026-04-19
No Icicles Block Tag Mixin

### LINKED ISSUE #3711: [💎]: Prevent icicles from generating under monorails
### Was this enhancement suggested before?

- [x] No, it was not.

### What would you like to be added/ changed?

Prevent icicles from generating under monorails

### Why is this needed?

using the monorail will just make me go out of my seat and throw me out of the monorail, + it can kill me + it's not very aesthetic , especially with curves monorails

Image

SakuraKitsurugi:
### What is the new behavior?
Mixins TFC's EnvironmentHelpers to add another check if a `NoIcicles` tag exists on a block.

### Additional Information
- This is my first Mixin so there might be issues.
- Mixin class for reference: https://github.com/TerraFirmaCraft/TerraFirmaCraft/blob/1.20.x/src/main/java/net/dries007/tfc/util/EnvironmentHelpers.java#L332
- Fixes https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3711


Pyritie (approved):

SakuraKitsurugi:
I tested it in the full modpack and it seems to crash.

Pyritie (approved):

Redeix (approved):

PR #432, opened 2026-04-20
Add TFC Anvil Machine

SpicyNoodle5:
Adds an electric anvil machine that performs TFC forging steps

### Future Work
- Add machine texture
- Verify forge bonuses can be applied
- Add a way to choose a specific anvil recipe when there are multiple options
- Restrict anvil recipe tier based on voltage
- Add recipes
- Maybe make a multiblock variant

**This is a work in progress, do not use this on worlds you care about.**


TomPlop:
I know it's been on stand by for quite some time but if you don't mind resolving conflicts once in a while so it's not a shit show if we decide to add it

Pyritie:
maybe we could make a branch and then you could PR to the branch? that way anyone could fix it up later

SpicyNoodle5:
This can probably just be closed tbh. The last version I worked on was a different branch anyway: https://github.com/SpicyNoodle5/Core-Modern/tree/anvil_machine_v2

PR #437, opened 2026-04-28
Create Fluid Logistics Migration

Bumperdo09:
## What is the new behavior?
Migrate to Create Fluid Logistics
For https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3727
TLDR Create Factory Logistics is quite buggy and no longer supported. Create Fluid Logistics generally works alot better

Also changes to Modpack-Modern to go along with this

## Implementation Details
Fluid packagers and fluid gauges are automatically migrated to their new versions when the chunk they are in is loaded for the first time.
- The fluid packager block and blockEntity is migrated using the forge remapIds event
- The fluid gauge block is migrated using the forge remapIds event
- The fluid gauge blockEntity is remapped using a mixin so that the blockEntity tag can be modified slightly to work with Fluid Logistics. Without the mixin modifying the tags the filters on the new gauges wont be set properly which results in breaking peoples setups.

## Outcome
When loading an old world the transition to the new mod should be seamless. From my testing using my current play through all my fluid logistics setups still work perfectly after the migration

## Additional Information
This shouldn't be merged until the next alpha release to limit the breaking of peoples worlds


SakuraKitsurugi (approved):

gradle/scripts/dependencies.gradle
Pyritie, line 1: we can just delete the commented lines here

Pyritie (approved):

PR #443, opened 2026-05-07
Fix gaswell

TomPlop:
-Fix the cache issue it could create
-Fix the #explosive not being consumed
-Add an active state to the Controller
-Switch Steam Input bus to Input Bus

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/common/tfgt/machine/trait/GasWellRecipeLogic.java
+++ b/src/main/java/su/terrafirmagreg/core/common/tfgt/machine/trait/GasWellRecipeLogic.java
@@ -38,6 +38,7 @@ public class GasWellRecipeLogic {
     public void reset() {
         timer = 0;
         hasConsumedExplosive = false;
+        machine.setActive(false);
     }
 
     public void resetFull() {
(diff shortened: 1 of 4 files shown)


src/main/java/su/terrafirmagreg/core/common/tfgt/machine/trait/GasWellRecipeLogic.java
```diff
@@ -96,6 +112,24 @@ public void tick() {
...
     }
 
+    private boolean consumeExplosive() {
+        var itemHandler = machine.getInputItemHandler();
+        if (itemHandler == null)
+            return false;
+
+        for (int i = 0; i < itemHandler.getSlots(); i++) {
+            var stack = itemHandler.getStackInSlot(i);
+            if (!stack.isEmpty() && isExplosive(stack)) {
```
Redeix, line 121: maybe im blind but where does isExplosive come from?

TomPlop: > maybe im blind but where does isExplosive come from?

It's a tag that is defined a bit further down

Redeix: oh gotcha, I totally was blind. I was worried it was some hardcoded item property lol. A tag is good

Redeix (approved):

gustovafing (approved):

PR #449, opened 2026-05-14
rotten ingredient wrapper ignores non-food items

Redeix:
If you use a notRotten condition on a non food item the stack just becomes null. Which is annoying. So now it just returns itself.


Pyritie (approved):

jmecn (approved):

Pyritie:
feel free to merge this whenever is appropriate

PR #460, opened 2026-05-21
Fix GT Solar Panels on mars checking the client's wind

### LINKED ISSUE #2665: [🐛]: Solar Panels on Mars Do Not Generate Power While Player Is in the Overworld

- [x] No, it was not.

### Version modpack used

0.11.16

### Environment

Local / Singleplayer (Integrated Server)

### New Worlds

Unsure

-

### Expected Behavior

Solar panels placed on Mars should generate power based solely on the environmental conditions of the Mars dimension (sunlight, planet rules).

Power generation should not depend on:

The player’s current dimension

The player’s position in the Overworld

Whether the player is on the surface or underground in another dimension

### Actual Behavior

Solar panels placed on Mars stop generating power when the player is located on the surface of the Overworld.

As soon as the player goes underground in the Overworld to a depth where the surface is no longer rendered, the solar panels on Mars immediately start generating power again.

In some cases, solar panels on Mars also do not work while the player is on the Overworld surface even when there is no precipitation or visible weather.

### Steps to Reproduce

1 Travel to Mars

2 Place and connect solar panels on Mars

3 Return to the Overworld and stay on the surface
   → Solar panels on Mars do not generate power

4 In the Overworld, go underground until the surface is no longer rendered
   → Solar panels on Mars start generating power

https://www.youtube.com/watch?v=Go0l44rFIjw

### Additional Information

Issue occurs on a local / integrated server

Reproducible consistently

Can be demonstrated (see pinned video)

Appears to be related to multi-dimensional logic

Possible causes:

Sunlight or sky checks using the player’s current dimension instead of the panel’s dimension

Energy generation logic depending on Overworld surface rendering state

Incorrect handling of dimension context for solar panels

ashleney:
Fixes https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/2665

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/GTUtilMixin.java
+++ b/src/main/java/su/terrafirmagreg/core/mixins/common/gtceu/GTUtilMixin.java
@@ -5,9 +5,9 @@ import org.spongepowered.asm.mixin.Overwrite;
 
 import com.gregtechceu.gtceu.utils.GTUtil;
 
-import net.dries007.tfc.client.ClimateRenderCache;
 import net.dries007.tfc.util.EnvironmentHelpers;
 import net.dries007.tfc.util.calendar.Calendars;
+import net.dries007.tfc.util.climate.Climate;
 import net.minecraft.core.BlockPos;
 import net.minecraft.world.level.Level;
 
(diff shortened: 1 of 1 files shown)


Redeix (approved):
Good catch

Pyritie (approved):

PR #461, opened 2026-05-22
Fix ad astra lander mounting not working because the chunk was presumably not loaded

ashleney:
Addresses https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3054

From what was shared about the issue and from what I've seen in the logs, this seems to be caused by failing to mount the Ad Astra lander because the chunk wasn't loaded. Due to the nature of this bug, I have no way to verify that this fix works.

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/mixins/common/ad_astra/ModUtilsMixin.java
+++ b/src/main/java/su/terrafirmagreg/core/mixins/common/ad_astra/ModUtilsMixin.java
@@ -8,10 +8,13 @@ import java.util.Objects;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Overwrite;
 
+import net.minecraft.core.BlockPos;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.server.level.TicketType;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.EntityType;
+import net.minecraft.world.level.ChunkPos;
 import net.minecraft.world.phys.Vec3;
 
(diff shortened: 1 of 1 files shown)


Pyritie (approved):
worth a shot I guess!

BlueBoat29:
yeah there is a good chance it does work. But since we aren't able to replicate the bug on our side its hard to say

SakuraKitsurugi:
From first idea it might not as well, I remember when I started playing and there were no double rockets - my friend was sitting in the base and I still had this issue happen when I went back.

ashleney:
This patch is not specific to double rockets so I'm not sure what you mean by that

SakuraKitsurugi:
I meant about the fact this fixing the issue because a chunk was not loaded. Even when loaded I had this issue in like 0.11.25.
 
 My entire base was chunk loaded.

Pyritie:
when were double rockets added? did this issue happen before then?

BlueBoat29:
people reported this happening back in 0.10, way before double rockets were added https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3054

ashleney:
After a lot of attempts on sakura's server, here's some ways we managed to get the issue (not at all reliable):

- launch to the moon
- launch to earth
- launch to the moon from a different spot
- launch to earth from a different spot, sneak while loading the world, there will be about a 1 in 3 chance you get stuck
- relog to land in a different spot

alternatively

- have a low tps server
- launch to a planet
- there's a small chance you get desynced and get out of the lander despite being in the lander on the expected destination server-side
- relog to land in a different spot

This makes it fairly improbable that loading will *actually* fix it but imo it's still worth trying.

SakuraKitsurugi (approved):

PR #465, opened 2026-05-24
Fix ownerless railgun crash

### LINKED ISSUE #3589: [💥]: Railgun creates a crash chunk when formed from schematicannon
### Was this crash reported before?

- [x] No, it was not.

No, I am not using Prism or Curseforge.

### Did you install required dependencies?

- [x] Not Applicable / Yes, I have all required dependencies installed.

### Computer OS

Win 10

### Modpack Version Used

0.11.26

### Minecraft Environment

Singleplayer

### Side Crashed

Client and Server

### New Worlds

Yes

### Please provide a link to your latest.log, if relevant

https://mclo.gs/IMHUSSh

### Steps to Reproduce

1. Create a new world
2. Place interplantary logistics machine (receiver or launcher) using Create schematicannon
3. Form a full multiblock from it
4. Get crash with "**java.lang.NullPointerException: Ticking block entity**"

World cannot be loaded in singleplayer as we effectively created a crash chunk

### Additional Information

Why this happens?

Create's schematicannon places blocks without OwnerUUID nbt value which is needed by [loadOrCreatePart](https://github.com/TerraFirmaGreg-Team/Core-Modern/blob/dev/src/main/java/su/terrafirmagreg/core/common/tfgt/interdim_logistics/InterplanetaryLogisticsNetwork.java#L70) method to add NetworkPart to Map<DimensionalBlockPos, NetworkPart> parts;

So when [getSendConfigurations](https://github.com/TerraFirmaGreg-Team/Core-Modern/blob/dev/src/main/java/su/terrafirmagreg/core/common/tfgt/interdim_logistics/InterplanetaryLogisticsNetwork.java#L140) is called it always fails to get NetworkPart as it was never added.

ashleney:
fixes https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3589

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/common/tfgt/interdim_logistics/InterplanetaryLogisticsNetwork.java
+++ b/src/main/java/su/terrafirmagreg/core/common/tfgt/interdim_logistics/InterplanetaryLogisticsNetwork.java
@@ -79,16 +79,20 @@ public class InterplanetaryLogisticsNetwork extends SavedData {
         boolean isReceiver = machine instanceof ILogisticsNetworkReceiver;
 
         var owner = machine.getMachine().getOwner();
+        UUID ownerId = null;
         if (owner instanceof FTBOwner ftbOwner) {
-            loadedMachines.put(machine.getDimensionalPos(), machine);
-            parts.computeIfAbsent(machine.getDimensionalPos(), k -> {
-                setDirty();
-                return new NetworkPart(k, ftbOwner.getTeam().getTeamId(), isReceiver);
-            });
-            return;
+            ownerId = ftbOwner.getTeam().getTeamId();
+        } else {
+            TFGCore.LOGGER.warn("Interplanetary logistics machine does not have a valid owner. {} {}",
+                    machine.getDimensionalPos(), machine.getMachine());
         }
-        TFGCore.LOGGER.warn("Interplanetary logistics machine does not have a valid FTB owner. {} {}",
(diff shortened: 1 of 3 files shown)


SakuraKitsurugi:
Hmm, I either don't quite get it, but how does this set the owner :thinking:

ashleney:
it doesn't, I essentially just added a bunch of graceful null checks instead of having requireNonNull

SakuraKitsurugi:
That... does not really sound like a good fix. :thinking:

ashleney:
What do you suggest? Giving it a fake owner id?

SakuraKitsurugi:
The fix probably has to also come from the schematicannon as well. Like checking who started the cannon and from that all blocks placed are owned by that player/team.

ashleney:
I suppose can look into how difficult that would be. I have concerns how stuff like turtles and deployers should be handled.

SakuraKitsurugi:
I think schematicanon should be main focus for now.

ashleney:
Schematicannon isn't at all compatible with gregtech's owner system. We would need a bunch of mixins just for the sake of the railgun. I'd rather add that in a separate PR if needed.

SakuraKitsurugi (approved):

SakuraKitsurugi:
Guess it is a good enough temporary solution. Maybe a better one will come up in the future.

Redeix (approved):

PR #468, opened 2026-05-26
Fix crowbars not being extractable from the create logistics network

ashleney:
unfortunately you can only select one crowbar of the same type at a time this way. I *think* I would need to add another mixin but I have frankly no idea how to do it properly. I tried a *lot* of stuff.


Pyritie:
only crowbars? would this not work to fix all gt tools?

src/main/java/su/terrafirmagreg/core/mixins/common/create/PackagerLinkBlockEntityMixin.java
```diff
@@ -0,0 +1,59 @@
...
+    private int tfg$allowGregTechToolRequests(InventorySummary summary, ItemStack stack) {
+        int exactCount = summary.getCountOf(stack);
+        if (exactCount > 0 || !(stack.getItem() instanceof IGTTool))
+            return exactCount;
+
+        return summary.getTotalOfMatching(other -> ItemStack.isSameItem(other, stack));
+    }
+
+    @WrapOperation(method = "processRequest", at = @At(value = "INVOKE", target = "Lcom/simibubi/create/content/logistics/packager/PackagingRequest;create(Lnet/minecraft/world/item/ItemStack;ILjava/lang/String;ILorg/apache/commons/lang3/mutable/MutableBoolean;IILcom/simibubi/create/content/logistics/stockTicker/PackageOrderWithCrafts;)Lcom/simibubi/create/content/logistics/packager/PackagingRequest;"), remap = false)
+    private com.simibubi.create.content.logistics.packager.PackagingRequest tfg$useMatchedGregTechToolStack(
```
Pyritie: why not put the namespace up in the imports?

ashleney:
ah well it's specifically crowbars that are the most common issue for players, stuff like hammers already worked.

src/main/java/su/terrafirmagreg/core/mixins/common/create/PackagerLinkBlockEntityMixin.java
ashleney: oops

Pyritie (approved):

Redeix (approved):

Robak132:
Sacks and similar stuff are also not extractable

ashleney:
@Robak132 Thanks for the info, I could only reproduce this for burlap sacks and those are fixed with the packager link mixin

Robak132:
> @Robak132 Thanks for the info, I could only reproduce this for burlap sacks and those are fixed with the packager link mixin

Ore sacks definitely have the same problem. I will check problematic items and go back to you in few hours.

Robak132:
Items that I find unextractable:
- leather sack
- burlap sack
- ore sack

ashleney:
yes I could reproduce this with sacks that have NBT data, I'll see what I can do to patch it

ashleney:
I believe the issue with sacks and crowbars not being extractable is now fixed. There's also another issue of some items that were specifically obtained with the cheat mode of EMI not being extractable but I have no idea how to fix that nor do I believe it particularly needs fixing on TFG's side.

Pyritie (approved):

Pyritie:
Someone on 0.13.6 reported they couldn't get their scythe out, is this expected? https://discord.com/channels/400913133620822016/1485640800586039316/1533023702457778347

PR #469, opened 2026-05-26
Fix dontBurnPassengers only checking the first passenger

### LINKED ISSUE #3317: [🐛]: 2 Player Rocket Occasionally Incinerates Secondary Rider

- [x] No, it was not.

### Computer OS

Windows 11

### Modpack Version Used

0.11.25

### Minecraft Environment

Forge Server

### New Worlds

Yes

n/a

### Expected Behavior

Rocket safely takes two players to their selected planet.

### Actual Behavior

Entering a two-person rocket with another player already in it can occasionally cause the player to burn up during flight, giving the "[Player] melted under an active rocket" death message. Occurs, from my experience, about 1/15 times, and has had no real noticeable cause. The player who entered the rocket second is always the one who dies.

### Steps to Reproduce

Enter rocket with another player already in it, launch. Chance you might burn up.

### Additional Information

This might be related to lag, but I wouldn't think so as the server I play on has almost zero lag issues.

ashleney:
*Might* fix https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3317, but I didn't actually check.

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/mixins/common/ad_astra/RocketMixin.java
+++ b/src/main/java/su/terrafirmagreg/core/mixins/common/ad_astra/RocketMixin.java
@@ -101,10 +101,9 @@ public abstract class RocketMixin extends Entity {
     @Redirect(method = "burnEntitiesUnderRocket", at = @At(value = "INVOKE", target = "net/minecraft/world/entity/LivingEntity.equals (Ljava/lang/Object;)Z"))
     private boolean tfg$dontBurnPassengers(LivingEntity instance, Object o) {
         List<Entity> passengers = tfg$self.getPassengers();
-
-        for (var entity : passengers) {
-            return instance.equals(entity);
-        }
+        for (var entity : passengers)
+            if (instance.equals(entity))
+                return true;
         return false;
     }
 


Pyritie:
we should probably actually check on a server

src/main/java/su/terrafirmagreg/core/mixins/common/ad_astra/RocketMixin.java
```diff
@@ -101,10 +101,9 @@ public RocketMixin(EntityType<?> pEntityType, Level pLevel) {
     @Redirect(method = "burnEntitiesUnderRocket", at = @At(value = "INVOKE", target = "net/minecraft/world/entity/LivingEntity.equals (Ljava/lang/Object;)Z"))
     private boolean tfg$dontBurnPassengers(LivingEntity instance, Object o) {
         List<Entity> passengers = tfg$self.getPassengers();
-
-        for (var entity : passengers) {
-            return instance.equals(entity);
-        }
+        for (var entity : passengers)
+            if (instance.equals(entity))
```
SakuraKitsurugi, line 105: From first glace it does not really feel like a solution because if this was the issue - the second passanger should always return false and start burning because the first one in the list is probably the first passenger. However, the second passenger does not _always_ burn.

SakuraKitsurugi: I've noticed that this can happen more often when lagging because on my test I would start off the rocket ride fine, but if I get a lag spike on any of my instances - the second player would start burning.

ashleney: even before this function was added the burning was nondeterministic. It's not the only thing preventing passengers from burning up.

TomPlop: I remember that GameStar added an immunity to fire to the second passenger when travelling. Honestly he's the best person to answer about this issue.

@BlueBoat29

Pyritie: I don't think he actually did, it was just a suggestion I made which he said might work as a workaround

SakuraKitsurugi: What about adding a check if passengers are entirely empty and also returning false instantly (maybe passengers do not get updated during lag?).

ashleney: The function already returns false if the passenger list is empty

BlueBoat29:
For context this is the method that I am mixing into.

`public void burnEntitiesUnderRocket() {
        if (level().isClientSide()) return;
        for (var entity : level().getEntitiesOfClass(LivingEntity.class, getBoundingBox()
            .inflate(2, 30, 2)
            .move(0, -37, 0), e -> true)) {
            if (entity.equals(getControllingPassenger())) continue;
            entity.setSecondsOnFire(10);
            entity.hurt(ModDamageSources.create(level(), ModDamageSources.ROCKET_FLAMES), 10);
        }
    }`

I believe that the redirect is targeting and replacing the entity.equals(getControllingPassenger())). So what we want is for the mixin to return true for any entity that shouldn't get set on fire.

I think you are right. While the original mixin has a for loop due to it just having a return, it will always exit on the first iteration, whether that is true or false. Originally, I thought this change wasn't needed, but after remembering how this system works, I can see that the original mixin is bugged. Good catch! 

As for why the second passenger is not always burned, my understanding is that is due to client server desync. Like the boundary box is set to be below the rocket and the rocket is travelling up. So if a player were to lag and the server thinks they are actually below the rocket (before their position is updated to match the rocket) they will get burned. This was prevented for the driver in the original code and my mixin due to them both essentially only checking the first passenger, which is the driver.

BlueBoat29 (approved):

SakuraKitsurugi (approved):
Huh, so it was partly due to lag.

Pyritie (approved):

PR #473, opened 2026-05-28
Swap Species dev patch for mixin

jurrejelle:
## What is the new behavior?
Swap out the patchSpeciesJar for a mixin that solves the species crash

## Outcome
Can launch a server without problems


Pyritie (approved):
the goat has arrived

Redeix:
I'll check it out tonight. Let's make Gustavo review it for the funny

gustovafing (approved):

PR #476, opened 2026-05-31
Slime Friends (Minimal Version)

SakuraKitsurugi:
## What is the new behavior?
Adds Slimes. Adds 1 basic slime, 3 producing slime and 2 ambiental slimes.

### Main
- [x] Main Class
  - [x] Basic Entity Setup
  - [x] Genderless Breeding
  - [x] Item Production
  - [x] Item Production Dimension Lock
  - [x] Pet Mechanics
  - [x] Ambiental Stats to Ice/Lava Slimes.
  - [x] Spawn Rules (Will need testing once merged)
- [x] Model
  - [x] Inner Body
  - [x] Outer Body Layer (Transparent)
  - [x] Face Layer
- [x] Items
  - [x] Datagen
  - [x] Slime Code Adjustment
- [x] Variants
  - [x] Variant Code

### Item Textures
- [x] Slime Ball (Multiple)

### Slime Textures
- [x] Plant
- [x] Glowberry
- [x] Spring
- [x] Ice
- [x] Lava
- [x] Latex

## Additional Information
- Making this a draft to keep progress and for others to review code earlier if I did any mistakes.

## Companion PRs:
- Modpack: https://github.com/TerraFirmaGreg-Team/Modpack-Modern/pull/4171
- Tools: https://github.com/TerraFirmaGreg-Team/Tools-Modern/pull/479


Redeix (approved):

src/main/java/su/terrafirmagreg/core/common/entity/slime/SlimeVariant.java
```diff
@@ -0,0 +1,88 @@
...
+        this.item = item;
+    }
+
+    // region Getters
+    public @NotNull String getSerializedName() {
+        return this.name;
+    }
+
+    public static SlimeVariant getByName(String name) {
+        if (variantNameMap.get(name) != null) {
```
Pyritie: this does a lookup twice

src/main/java/su/terrafirmagreg/core/common/entity/slime/TFGSlime.java
```diff
@@ -0,0 +1,344 @@
...
+        this.setVariant(SlimeVariant.getByName(tag.getString("variant")));
+        this.setProducedTick(tag.getLong("produced"));
+    }
+
+    public SpawnGroupData finalizeSpawn(ServerLevelAccessor level, DifficultyInstance difficulty, MobSpawnType type, @Nullable SpawnGroupData data, @Nullable CompoundTag tag) {
+        SpawnGroupData spawnData = super.finalizeSpawn(level, difficulty, type, data, tag);
+        this.setGender(Gender.FEMALE);
+        this.setBaby(this.random.nextFloat() < 0.1F);
+        this.setVariant(this.initialVariant(level));
+        return super.finalizeSpawn(level, difficulty, type, spawnData, tag);
```
Pyritie: why not just return `spawnData` here instead of calling `super.finalizeSpawn()` twice?

src/main/java/su/terrafirmagreg/core/common/entity/slime/TFGSlimeFaceLayer.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.client.renderer.entity.RenderLayerParent;
+import net.minecraft.client.renderer.entity.layers.RenderLayer;
+import net.minecraft.resources.ResourceLocation;
+
+import su.terrafirmagreg.core.TFGCore;
+
+@SuppressWarnings({ "unchecked" })
+public class TFGSlimeFaceLayer extends RenderLayer<TFGSlime, TFGSlimeModel<TFGSlime>> {
+    public static final ModelLayerLocation LAYER_LOCATION = new ModelLayerLocation(
+            ResourceLocation.fromNamespaceAndPath(TFGCore.MOD_ID, "slime_face"), "main");
```
Pyritie: can use `TFGCore.id`

src/main/java/su/terrafirmagreg/core/common/entity/slime/TFGSlimeModel.java
```diff
@@ -0,0 +1,77 @@
...
+import net.minecraft.client.model.geom.ModelPart;
+import net.minecraft.client.model.geom.PartPose;
+import net.minecraft.client.model.geom.builders.*;
+import net.minecraft.resources.ResourceLocation;
+
+import su.terrafirmagreg.core.TFGCore;
+
+public class TFGSlimeModel<T extends TFGSlime> extends HierarchicalModel<T> {
+    public static final ModelLayerLocation LAYER_LOCATION = new ModelLayerLocation(
+            ResourceLocation.fromNamespaceAndPath(TFGCore.MOD_ID, "slime"), "main");
```
Pyritie: can use `TFGCore.id`

src/main/java/su/terrafirmagreg/core/common/entity/slime/TFGSlimeOuterLayer.java
```diff
@@ -0,0 +1,48 @@
...
+import net.minecraft.client.renderer.entity.LivingEntityRenderer;
+import net.minecraft.client.renderer.entity.RenderLayerParent;
+import net.minecraft.client.renderer.entity.layers.RenderLayer;
+import net.minecraft.resources.ResourceLocation;
+
+import su.terrafirmagreg.core.TFGCore;
+
+public class TFGSlimeOuterLayer extends RenderLayer<TFGSlime, TFGSlimeModel<TFGSlime>> {
+    public static final ModelLayerLocation LAYER_LOCATION = new ModelLayerLocation(
+            ResourceLocation.fromNamespaceAndPath(TFGCore.MOD_ID, "slime_outer"), "main");
```
Pyritie: can use `TFGCore.id`

Pyritie (changes requested):

src/main/java/su/terrafirmagreg/core/common/entity/slime/SlimeVariant.java
SakuraKitsurugi: Not sure how to fix this one I will be honest, I even asked Mqrius :sob:

src/main/java/su/terrafirmagreg/core/common/entity/slime/TFGSlime.java
SakuraKitsurugi: I will be honest, just was a bit confused. Looking at TFC's code they do it randomly everywhere. Some do it like that, some just return `spawnData` and some just return the super at the end without calling it into a variable.

SakuraKitsurugi: However, fixed it.

src/main/java/su/terrafirmagreg/core/common/entity/slime/SlimeVariant.java
SakuraKitsurugi: Wait I am dumb, maybe like this?

src/main/java/su/terrafirmagreg/core/common/data/TFGEntities.java
Pyritie, line 1: does this actually solve the problem you were having? doesn't this make them spawn in the air?

src/main/java/su/terrafirmagreg/core/common/entity/slime/TFGSlime.java
```diff
@@ -153,9 +153,26 @@ public static AttributeSupplier.Builder createAttributes() {
...
+        if (level.getBlockState(cursor.below()).isAir()) {
+            return false;
+        }
+
+        if (Helpers.isFluid(level.getFluidState(cursor), FluidTags.LAVA)
+                || Helpers.isFluid(level.getFluidState(cursor.below()), FluidTags.LAVA)) {
+            return false;
+        }
+
+        while (!level.getFluidState(cursor).isEmpty()) {
```
Pyritie: what does this do?

SakuraKitsurugi: similar to how elks spawn, if it detects that it spawns in a liquid - it moves the spawnpoint above it

src/main/java/su/terrafirmagreg/core/common/data/TFGEntities.java
SakuraKitsurugi: It still checks Slime's spawn rules which do not let spawning in air.

src/main/java/su/terrafirmagreg/core/common/entity/slime/TFGSlime.java
Pyritie: how does it do that? the cursor isn't returned anywhere

I wouldn't copy red elks because they constantly spawn in the middle of the air and the only reason you don't see random venison on the ground any more is because I had to make them immune to fall damage with a mixin

SakuraKitsurugi: now that you say it, I don't exactly know. Elk code is like this, but even then at the end its just a boolean check:
```java
public static boolean spawnRules(EntityType<? extends NetherPrey> type, LevelAccessor level, MobSpawnType spawn, BlockPos pos, RandomSource rand)
    {
        final BlockPos.MutableBlockPos cursor = pos.mutable();
        do
        {
            cursor.move(0, 1, 0);
        }
        while(Helpers.isFluid(level.getFluidState(cursor), FluidTags.LAVA));

        return level.getBlockState(cursor).isAir();
    }
```

SakuraKitsurugi: I don't actually know what is the difference between just `while` and `do` + `while`

SakuraKitsurugi: anyway, changed it to a simple check

Pyritie (approved):

src/main/java/su/terrafirmagreg/core/common/entity/slime/TFGSlime.java
Pyritie: `while` will check the condition first before doing anything in the loop, whereas `do + while` will always execute the loop contents at least once, and then check the condition before doing it again

PR #439, opened 2026-04-30
Fix TFCGenViewer new worldgen

Thomasx0:
Closes issue TerraFirmaGreg-Team/Modpack-Modern#3834
A bit clunky but works

PR #484, opened 2026-06-05
Fix cindercones again

### LINKED ISSUE #4140: [🐛]: Cinder cones aren't always covered in basalt

- [x] No, it was not.

### Computer OS

N/A

### Modpack Version Used

0.12.8

### Minecraft Environment

Singleplayer

### New Worlds

Yes

N/A

### Expected Behavior

All cinder cones should have basalt on them

### Actual Behavior

Cinder cones do not always have basalt on them

### Steps to Reproduce

7376638828777845309 @ 3050 ~ 12900
7376638828777845309 @ -640 ~ -1035
-1164348183722065023 @ 12517 ~ -4715

### Additional Information

I swear we fixed this back in 0.12.5 or so.
This should be fixed both for tfc consistency but also because it looks odd having lava right next to grass, and it often starts large forest fires.

The cause is something to do with the basalt-applying "step" not always generating at the right heights? Or maybe not at all? At least it doesn't seem to be misaligned like it used to be.

If we can't figure out what causes this, I can probably "fix" it with a very large disc feature.

Mqrius:
Fixes https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/4140

## Bug cause:
The bug was introduced by #439, which loads biomes early (during mod loading I believe) to register map colors. At this time of loading, the worldgen version hasn't been set yet. Thomas made a workaround for it by inferring the worldgen version and/or accounting for all edge cases, but the workaround wasn't applied to the surfacebuilders. So the wrappers never got installed.

A deeper issue is that the `OVERWORLD_VERSION` uses `0` both to mean "not initialized" and to mean "old style worldgen". This makes bugs more likely and makes them harder to trace. _(Who coded this? Ah, me from a few months ago! Such an idiot)_

## Fix:
I removed the overworld version gate from the surface builders, and added a null check where needed. Now, the surface builders will always exist, but they exit out if the biome doesn't exist or if the BiomeExtension doesn't have tfg$hasCinderCones set.

I think this should be fine in all cases, and it definitely works in the new worldgen new world case, but I haven't actually checked if there's any issues if you use the old worldgen. Might be worth checking if anyone still runs an old world.

## Alternative fix possible but not implemented:
I believe `BiomeColorsMixin.tfg$preRegisterTfgBiomeDataSlots` doesn't need to run at `<init>`, instead it could run at `apply()` or lazily in `color()`. That should avoid the issues with loading the biomes too early as well, while still registering the colors for the map, and not needing to infer any worldgen versions.


Pyritie:
out of curiosity, did you check tuyas and tuff rings, since they also use the same system?

Pyritie:
@Thomasx0

SakuraKitsurugi:
Could this partly fix another issue?: https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3950

Mqrius:
> out of curiosity, did you check tuyas and tuff rings, since they also use the same system?

I haven't checked, but from what I saw they would have been broken as well, and would be fixed by this fix as well.

> Could this partly fix another issue?

Uhhh I don't think this patch affects that. The deeper root cause seems related though; the worldgen version we use should be a property of a generated world, but we need to know it before the world is generated, which makes things super messy and is partly responsible for these bugs. I haven't looked closely at what Thomas did with the `TfgClientPreviewState.useTfgOverworldPipeline()`, but that's probably the place I'd look at for the issue you linked.

Pyritie:
given that you can't create worlds with the old worldgen and that screen is only incorrect at world creation, forcing the new biome selection at that point would be an ok compromise I think?

Pyritie (approved):

SakuraKitsurugi (approved):

Thomasx0:
Sorry, I had to leave for work for almost 2 months (no access to my home PC :*( ), I'll be back around the 20th. Issue [#3950](https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3950) is ready, and I've also finished [#3865](https://github.com/TerraFirmaGreg-Team/Modpack-Modern/issues/3865). But it will need testing

Pyritie:
ooh sounds good!

PR #489, opened 2026-06-09
Mining restrictions

Pyritie:
- Tells miners that their minimum Y value they can mine is 80 in the beneath
- Contraptions breaking blocks below a Y level in the beneath that would return ore items, have those ore items removed (so they just return tiny dusts and stone dust - other blocks behave as normal)
- Similar behaviours for venus and mercury
- Multiblock miners behave as normal
- Config options for all of the above just cuz why not
- Also a mixin to make ghouls shut up because they're way too noisy


src/main/java/su/terrafirmagreg/core/mixins/common/create/MovementBehaviourMixin.java
```diff
@@ -0,0 +1,67 @@
...
+            var dim = context.world.dimension();
+
+            // Don't drop ores below Y=80
+            if (dim == Level.NETHER && TFGConfig.SERVER.enableBeneathMiningRestrictions.get()) {
+                if (context.position.y <= TFGConfig.SERVER.disabledBeneathMiningYLevel.get()) {
+                    return;
+                }
+            }
+            // Don't drop ores at all on venus/mercury
+            else if ((dim == Planet.VENUS || dim == Planet.MERCURY) && TFGConfig.SERVER.enableHotPlanetMiningRestrictions.get()) {
```
Redeix, line 42: :eyes:

Redeix (approved):

Pyritie:
btw I picked y=80 because it's about where the bottom layer starts, and is also where the current temp = 40C, which feels like a good spot.

TomPlop (approved):

PR #496, opened 2026-06-12
migrate tfc ambiental to Core-Modern

ashleney:
main differences:
- the base local temperature now directly corresponds to tfc climate
- you no longer check every single block twice when checking temperature
- you no longer get up to 4C of change from a good diet
- some changes to how block temperature is counted means that being surrounded by ice blocks will make you even colder than before
- added `/tfg ambientalDump` for debugging
- you're no longer cozy while underwater
- you don't have to be outside for cold blocks to affect you
- sea ice, mars ice and dry ice now make you cold
- snow no longer makes you cold even if you have shoes on, as originally intended
- water particles now have a configurable 30% chance to spawn per tick instead of a constant 50%


Pyritie:
I see you changed the namespaces, does this still work with tfc textile? it has ambiental as a dependency

ashleney:
it does now

src/main/resources/assets/tfcambiental/lang/en_us.json
Pyritie, line 1: these should be over in tools-modern instead

src/main/resources/data/tfcambiental/recipes/crafting/burlap_cowl.json
Pyritie, line 1: for these crafting recipes, let's see if any of them are overridden in the modpack and delete the files here. I know the straw hat is one of them

src/main/java/su/terrafirmagreg/tfcambiental/TFCAmbientalConfig.java
```diff
@@ -0,0 +1,185 @@
...
+
+            noiseArea = builder
+                    .comment("How big should noise areas be?")
+                    .defineInRange("noiseArea", 10, 3, 50);
+
+            drippiness = builder
+                    .comment("How much to drip when wet. 0 to turn off")
+                    .defineInRange("drippiness", 30, 0, 100);
+
+            seasonColorSummer = builder
```
Pyritie: might be good to clean up some of the configs in here that aren't used at all (I wonder what these ones were even for?)

src/main/java/su/terrafirmagreg/tfcambiental/item/Snowshoes.java
Pyritie, line 1: delete these snowshoes since ambiental never finished them and we have our own that thomasx0 made

Pyritie (changes requested):

src/main/java/su/terrafirmagreg/tfcambiental/item/Snowshoes.java
ashleney: unfortunately textile still needs some snowshoe-related stuff

Pyritie:
idk what you keep doing but whenever you do these force pushes it loses track of where my review progress was :')

src/main/resources/data/tfcambiental/recipes/crafting/burlap_cowl.json
Pyritie: straw hat recipe is still here, that's a knapping recipe in the modpack

src/main/java/su/terrafirmagreg/tfcambiental/item/Snowshoes.java
Pyritie: annoying - let's add a comment on the snowshoe files to remind ourselves though, like "only included because tfc textile requires it"

Pyritie (commented):
.

Pyritie (approved):

Pyritie:
should we wait to merge until after the next release?

ashleney:
I think this port could also use some testing, so I'd personally appreciate if it made it into this next alpha

Pyritie:
I wasn't planning on the next release being an alpha

ashleney:
I do not believe there's anything that could cause issues since this is mostly just a migration and bugfix PR, so imo it should be fine to merge but it's up to your discretion.

Pyritie:
I suppose it doesn't really matter too much, and yeah we'll likely need to do a hotfix right after anyway

Redeix (approved):
Gustavo is going to hate that it all still uses deferred register lol

PR #499, opened 2026-06-16
fix: adds missing tags to the new tfg wood chests

hollow-earth:
## What is the new behavior?
Added the following to tfg's new chests and trapped chests:
- tfg:default_chests (item tag)
- forge:chests/wooden (item tag)
- forge:chests/wooden (block tag)

## Outcome
See TerraFirmaGreg-Team/Modpack-Modern#4260
The new tags are registered in game and the inventory limit (18 items currently) is now shown.

## Additional Information
~~I was unable to add the `alekiships:can_place_in_compartments` and `firmaciv:chests` tags.~~ Tags were added.

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/common/data/blocks/TFGBlocks_Wood.java
+++ b/src/main/java/su/terrafirmagreg/core/common/data/blocks/TFGBlocks_Wood.java
@@ -502,8 +502,8 @@ public class TFGBlocks_Wood {
                     prov.simpleBlock(ctx.getEntry(), prov.models().getBuilder(ctx.getName()).texture("particle", wood.plankTexture));
                 })
                 .addLayer(() -> RenderType::cutout)
-                .tag(TagKey.create(Registries.BLOCK, ResourceLocation.fromNamespaceAndPath("forge", "chests")))
-                .tag(TagKey.create(Registries.BLOCK, ResourceLocation.fromNamespaceAndPath("forge", "chests/wooden")))
+                .tag(TagKey.create(Registries.BLOCK, ResourceLocation.fromNamespaceAndPath("forge", "chests")),
+                        TagKey.create(Registries.BLOCK, ResourceLocation.fromNamespaceAndPath("forge", "chests/wooden")))
                 .tag(BlockTags.MINEABLE_WITH_AXE)
                 .onRegister(block -> {
                     TFGBlockEntities.addValidBEBlock(TFCBlockEntities.CHEST, block);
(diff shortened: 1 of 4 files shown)


src/main/java/su/terrafirmagreg/core/common/data/blocks/TFGBlocks_Wood.java
```diff
@@ -512,7 +513,10 @@ private static BlockEntry<Block> chest(TFGWood wood) {
                     prov.withExistingParent(ctx.getName(), ResourceLocation.withDefaultNamespace("item/chest"))
                             .texture("particle", wood.plankTexture);
                 })
-                .tag(TagKey.create(Registries.ITEM, ResourceLocation.fromNamespaceAndPath("forge", "chests"))).build()
+                .tag(TagKey.create(Registries.ITEM, ResourceLocation.fromNamespaceAndPath("forge", "chests")))
```
SakuraKitsurugi: You do not need to create a new tag for all of these. Forge tags (at least def simple ones like forge:chests) already have them defined that you can use. Also you can combine `.tag` into one by simply adding commas.

SakuraKitsurugi (changes requested):

src/main/java/su/terrafirmagreg/core/common/data/blocks/TFGBlocks_Wood.java
hollow-earth: Sure, I can refactor that, no worries. However, this is the smallest subset of tags I could find that matched the tags that are currently used for the TFC chests. I'll push a new commit in a moment.

Ash chests (TFC):
Image

TFG chests pre-commit
Image

TFG chests after adding forge:chests/wooden (item and block tags)
Image

TFG chests after adding tfg:default_chests:
Image

And finally, TFG chests after adding firmaciv:chests:
Image

Pyritie (approved):

Pyritie:
I want to dig into this myself a bit and find out why some of these tags aren't being automatically applied or why carryon can't see them

SakuraKitsurugi (approved):
I will fix it up later after my university stuff, for now I will approve.

hollow-earth:
> I want to dig into this myself a bit and find out why some of these tags aren't being automatically applied or why carryon can't see them

@Pyritie so far, \#forge/chests, forge/chests, \#forge:chests/wooden, and forge:chests/wooden seem to have no effect in the carryon config. Tried some variations with * to no avail.

PR #505, opened 2026-06-22
Make items/containers consume more energy with the interplanetary card

ashleney:
not too sure about the numbers


TomPlop (commented):
Not sure about the numbers either but why make it cheaper for whitelisted items?

For blacklisted items having it multiplied by 4 doesn't seem enough as it doesn't really solve the issue I think

TomPlop (approved):

TomPlop:
Could you add a tooltip to let the player knows please

https://github.com/TerraFirmaGreg-Team/Modpack-Modern/blob/17b8350553a95bcc30be71c88d2f5356ac011cf1/kubejs/client_scripts/tooltips.js#L516

Pyritie (approved):

PR #511, opened 2026-06-27
Make player temperature accurate inside the EBF

ashleney:
When the player crawls into any multiblock that uses coils they will heat up to the coil temperature while the multiblock is active. If they crawl into a vacuum freezer they'll cool down.

Additionally:
- changed heatstroke/frostbite behavior so that while above 30C instead of dealing 2 hearts every ~43s it now deals **half a heart every ~2s**, increasing every additional 15C. The previous behavior was probably a bug.
- potency is no longer set to 1 when over 25C, instead it's divided by 4. Originally it always took at least ~11 minutes to heat up from 25C to 30C, this makes heating up much faster when inside a multiblock for example. For most cases (such as exploring the beneath) this is effectively unchanged.

### DIFF
--- a/src/main/java/su/terrafirmagreg/tfcambiental/api/BlockTemperatureProvider.java
+++ b/src/main/java/su/terrafirmagreg/tfcambiental/api/BlockTemperatureProvider.java
@@ -14,8 +14,13 @@ import java.util.stream.Stream;
 import com.eerussianguy.firmalife.common.blocks.OvenBottomBlock;
 import com.gregtechceu.gtceu.api.block.property.GTBlockStateProperties;
 import com.gregtechceu.gtceu.api.capability.GTCapabilityHelper;
+import com.gregtechceu.gtceu.api.machine.IMachineBlockEntity;
+import com.gregtechceu.gtceu.api.machine.multiblock.CoilWorkableElectricMultiblockMachine;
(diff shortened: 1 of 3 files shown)


Pyritie (approved):

src/main/java/su/terrafirmagreg/tfcambiental/api/BlockTemperatureProvider.java
```diff
@@ -358,4 +364,28 @@ static Optional<TempModifier> getBlockTempModifier(Player player, BlockPos block
 
         return Optional.empty();
     }
+
+    static Optional<TempModifier> handleMultiblockInside(Player player, BlockEntity blockEntity) {
+        if (blockEntity instanceof IMachineBlockEntity machineBE &&
+                machineBE.getMetaMachine() instanceof WorkableElectricMultiblockMachine machine &&
+                machine.isFormed()
+                && machine.getMultiblockState().isPosInCache(player.blockPosition())
```
Redeix, line 372: Wow I didnt expect it to be that easy. Proper APIs are cool lol

Redeix (approved):

src/main/java/su/terrafirmagreg/tfcambiental/api/BlockTemperatureProvider.java
ashleney: me neither lol, I wrote a few mixins before realizing I can just do this

PR #515, opened 2026-06-29
fix #4304: Beneath custom spawn login/respawn race 

Thomasx0:
Fixes Beneath spawn being re-applied on relog/death. Defers first-join teleport until custom spawn is resolved, respects bed respawn, and leaves Overworld spawn presets untouched.


jmecn (approved):

BlueBoat29 (approved):
We will want to do server testing, but it looks good

PR #519, opened 2026-07-06
Misc ambiental changes

ashleney:
I removed the guinea bee for handleCozy and instead used a basic BFS. I tried keeping the bee but it behaved very unpredictably in many scenarios

changelog:
- temperature change speed has been temporarily halved until there's a more reasonable way to fight against extreme temperatures.
- lowered the temperature of many blocks. Players should no longer reach 100C when next to an EBF or -20C when surrounded by snow.
- temperature damage has been lowered significantly and no longer makes the player lose thirst/hunger
- when in a cold climate, small completely enclosed rooms will significantly warm the player up. Smaller rooms are considered warmer.
- encased fans from Create now slightly cool the player down


Pyritie:
I wonder if mqrius' new system could be a replacement for isCosy? since it's also an algorithm to find enclosed areas?

ashleney:
I think it can once it gets merged. Right now it's a bit too complex to just copy over to ambiental.

Redeix (approved):

TomPlop:
So basically right now we have:

Temperature increases again like it was doing before the fix
Most sources of heat are lower
Create Fan can decrease temperature

For the new room system I wonder if it can't be really expensive tps wise? Is there some ways we could make it better? Check a certain amount of blocks every ticks rather that all 20k? Also could we use it so small rooms are cozier that bigger ones? Right now a cave and a tiny house would bring the same values?

TomPlop:
Oh okay I didn't read Pyritie comment, I guess that makes sense though 20k block still seems really expensive tps wise?

ashleney:
I'll take a look at it

ashleney:
reverted making warehouses cozy, made it so that tiny rooms are warmer.

TomPlop (approved):

PR #528, opened 2026-07-13
Fix Ad Astra Lander

Black0305:
Fix for https://github.com/orgs/TerraFirmaGreg-Team/projects/10/views/1?pane=issue&itemId=156729680&issue=TerraFirmaGreg-Team%7CModpack-Modern%7C3054

## What is the new behavior?

Fixes a hard freeze when arriving on a planet with an Ad Astra lander on dedicated servers. After landing, the player is mounted on a lander the client never receives, so they are stuck: movement is dead (only head pitch works), dismounting does nothing, the lander menu has 0 slots, and the only recovery is a relog.

Root cause: `ModUtils.land()` (which this mod already `@Overwrite`s) spawns the lander and mounts the player in the **same tick** as the dimension change. On a dedicated server the player's entity tracker in the destination level isn't established yet at that point, so the lander's spawn packet is never sent to the client — the player ends up riding a "phantom" vehicle.

## Implementation Details

The only change is in `ModUtils.land()`: instead of calling `startRiding` immediately after `addFreshEntity`, the mount is deferred to a later server tick via a `TickTask` (`server.getTickCount() + 10`), guarded by `isAlive()` and `getVehicle() == null` checks.

- Deferring the mount lets the entity tracker send the lander to the client before the passenger packet, so the client actually has the vehicle when it mounts.
- 10 ticks (~0.5 s) is deliberately generous to give slower/laggier servers headroom.
- The `isAlive()` / `getVehicle() == null` checks avoid mounting a discarded lander or double-mounting.
- The player is briefly unmounted (~0.5 s) and free-falls before snapping onto the lander, which falls with them, so it isn't noticeable in practice.

## Outcome

Fixes the "stuck riding an invisible lander / must relog after landing on a planet" bug on dedicated servers.
Tested on my own server.

Minimal. The change is confined to the already-overwritten `ModUtils.land()`, adds no API or registry changes, and only shifts *when* the mount happens by a few ticks.

**Discord:** mixymixery

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/mixins/common/ad_astra/ModUtilsMixin.java
+++ b/src/main/java/su/terrafirmagreg/core/mixins/common/ad_astra/ModUtilsMixin.java
@@ -9,6 +9,7 @@ import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Overwrite;
 
 import net.minecraft.core.BlockPos;
+import net.minecraft.server.TickTask;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.server.level.TicketType;
@@ -54,7 +55,11 @@ public class ModUtilsMixin {
 
                     lander.setPos(pos);
                     targetLevel.addFreshEntity(lander);
-                    teleportedPlayer.startRiding(lander);
+                    targetLevel.getServer().tell(new TickTask(targetLevel.getServer().getTickCount() + 10, () -> {
+                        if (teleportedPlayer.isAlive() && lander.isAlive() && teleportedPlayer.getVehicle() == null) {
+                            teleportedPlayer.startRiding(lander, true);
+                        }
+                    }));
 
                     if (player != pilotPlayer)
                         continue;


Redeix (approved):

Pyritie:
@Black0305 hi, could we get confirmation that these contributions were made without AI?

BlueBoat29:
I’m more curious if you have been able to get a testing scenario where you can reliably reproduce the bug. Despite how often it’s reported nobody on the dev team has been able to reproduce it enough to know if any fix actually fixed the bug

SakuraKitsurugi:
The only way me and @ashleney were _able_ to reproduce it was getting  the server to a literal crawl (2 tps), but even then it was barely reliable.

ashleney:
it happened on 20tps in some circumstances tbf

Black0305:
It took me nearly a whole day to troubleshoot this even with AI.
Unfortunately my test setup is my own server where we play together.
Server is on 20 TPS. Our Moon landingspot is only one chunk off from the starting point on earth. Flying to the moon and back works flawless. But the Mars base is 800 Blocks north from the first landing spot. Flying to mars works most of the time but coming back to earth never works. I get the freeze and after a relog I´m on the coordinates where I left mars, but on earth. Sucks to swim hose 800 blocks back to the base.
Screenshot 2026-07-14 171308
Screenshot 2026-07-14 171317
Screenshot 2026-07-14 171323

SakuraKitsurugi:
> It took me nearly a whole day to troubleshoot this even with AI. Unfortunately my test setup is my own server where we play together. Server is on 20 TPS. Our Moon landingspot is only one chunk off from the starting point on earth. Flying to the moon and back works flawless. But the Mars base is 800 Blocks north from the first landing spot. Flying to mars works most of the time but coming back to earth never works. I get the freeze and after a relog I´m on the coordinates where I left mars, but on earth. Sucks to swim hose 800 blocks back to the base. Screenshot 2026-07-14 171308 Screenshot 2026-07-14 171317 Screenshot 2026-07-14 171323

I don't quite get it. Does that mean that this fix does not entirely fix the issue? Or are you explaining replication steps?

Black0305:
Those were my replication steps. With the patch everything works flawless when coming back to earth from mars

Black0305:
But this only happens on the server for me. In Singleplayer I couldnt replicate the freeze.

TomPlop (approved):

TomPlop:
Llm used but small tweaks

Be wary about this fix on the long term though, unknown if it fixes every issues.

Please refrain to use Llm or worst not advised you used them

PR #540, opened 2026-07-28
Moving Sound to Datagen + Vulcanizer & Seal Sounds

SakuraKitsurugi:
## What is the new behavior?
Moves all of our sounds to datagen so we do not need to manually write `sounds.json`. Additionally, I added the geologic vulcanizer and seal sounds.

## Additional Information
- Due to the way GT's registrate works - this method is limited. If volume or pitch needs to be changed for TFC sounds, it has to be done manually to the files.


src/generated/resources/assets/tfg/sounds.json
Pyritie, line 1: kubejs has its own file in this same location, which may cause issues - the kubejs one might need to be moved? I use it for music stuff

src/main/java/su/terrafirmagreg/core/common/data/TFGSounds.java
```diff
@@ -3,46 +3,41 @@
...
+    public static final SoundEntry BISON_HURT = TFGCore.REGISTRATE.sound(TFGCore.id("bison_hurt")).playExisting(TFCSounds.DEER.hurt(), 1.0f, 1.0f).build();
+    public static final SoundEntry BISON_WALK = TFGCore.REGISTRATE.sound(TFGCore.id("bison_walk")).playExisting(SoundEvents.WOLF_STEP).build();
+    public static final SoundEntry BISON_ATTACK = TFGCore.REGISTRATE.sound(TFGCore.id("bison_attack")).playExisting(TFCSounds.MOOSE.attack().get(), 1.0f, 1.0f).build();
+
+    public static final TFCSounds.EntitySound SEAL = new TFCSounds.EntitySound(SEAL_AMBIENT::getMainEvent, SEAL_DEATH::getMainEvent, SEAL_HURT::getMainEvent, SEAL_WALK::getMainEvent,
+            Optional.of(SEAL_ATTACK::getMainEvent), Optional.empty());
+    public static final TFCSounds.EntitySound BISON = new TFCSounds.EntitySound(BISON_AMBIENT::getMainEvent, BISON_DEATH::getMainEvent, BISON_HURT::getMainEvent, BISON_WALK::getMainEvent,
+            Optional.of(BISON_ATTACK::getMainEvent), Optional.empty());
     public static final TFCSounds.EntitySound FOX = new TFCSounds.EntitySound(() -> SoundEvents.FOX_AMBIENT, () -> SoundEvents.FOX_DEATH, () -> SoundEvents.FOX_HURT, () -> SoundEvents.CHICKEN_STEP,
```
Pyritie, line 33: can the tamed fox not just use the normal tfc fox sounds? or is it a bit different

SakuraKitsurugi: It is a bit different, TFCFox does not use TFCSounds at all.

src/generated/resources/assets/tfg/sounds.json
SakuraKitsurugi: I am always more for keeping stuff in Core, but I think it doesn't get overwritten? I will test it.

Pyritie: there's two sounds.json files in kjs - one in minecraft (has music for the overworld and menu) and one in tfg (has music for all the other dims). The tfg one might be able to be merged into the minecraft one

SakuraKitsurugi: Not necessary at least from my testing. Both sounds from kubejs and Core get merged.

Pyritie: interesting, that's inconsistent with other kinds of resourcepack behaviour, but I'm not gonna complain lol

Pyritie: did you check that the music and ambient sounds are still ok? like if you go to the moon you should hear some warped forest sounds

SakuraKitsurugi: It does :+1:

Pyritie (approved):

gustovafing (approved):

PR #544, opened 2026-07-30
Mixin to enable stencil texture at an early stage

Vazde:
## What is the new behavior?
Basicly does what state setup `Minecraft.getInstance().getMainRenderTarget().enableStencil();` does, but at a very early stage.

Should fix TACZ scopes and nutrients screen from corrupting Distant Horizons far chunk rendering.

## Outcome
Hovering the nutrients screen or equiping a scoped weapon should no longer break rendering.

Please test this thorougly. I didn't expect this to really work.

## Analysis
The root cause of this bug can be said to be Distant Horizons itself. Since Forge/Minecraft exposes an API which can mutate rendering state at runtime, DH should have an event listener so that it can invalidate its own cache. Should such an event API be missing, it is a shortfall with Forge, and more care and documentation needs to appear on the relevant state reading APIs to note what is invariant and what is not.

### DIFF
--- a/src/main/java/su/terrafirmagreg/core/mixins/client/minecraft/RenderTargetStencilMixin.java
+++ b/src/main/java/su/terrafirmagreg/core/mixins/client/minecraft/RenderTargetStencilMixin.java
@@ -1,0 +1,25 @@
+package su.terrafirmagreg.core.mixins.client.minecraft;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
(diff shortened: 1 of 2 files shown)


Pyritie (approved):
haha what's a testing

SakuraKitsurugi (approved):

Vazde:
I just noticed that this might not be the ultimate solution for some reason. My monitor went to power save mode, and once resuming the far chunks got broken. Even without hovering the nutrition graph or a scoped weapon. The mixin should have prevented this as it patches the stencil at buffer at create time. But at least it should improve things. That may also have been something else unrelated.

PR #546, opened 2026-08-01
Earth Jellie Spawns

SakuraKitsurugi:
## What is the new behavior?
Sets biomes that earth jellies spawn in.

## Additional Information
I honestly have no idea if these locations are in any way good.

## Companion PRs
- Modpack: [WIP]

### DIFF
--- a/gradle/scripts/dependencies.gradle
+++ b/gradle/scripts/dependencies.gradle
@@ -177,7 +177,7 @@ dependencies {
     modCompileOnly "curse.maven:additional-placements-674852:7905535"
 
     // Jellies
-    modImplementation "maven.modrinth:jellies:0.1.2"
+    modImplementation "maven.modrinth:jellies:0.1.3"
 
     // Create: Steam Powered, 22 jul 2026
     modImplementation "maven.modrinth:create-steam-powered:1.20.1-3.1.0"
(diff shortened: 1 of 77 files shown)


Pyritie (changes requested):
mm, these are too few. If we want them to be in different biomes, we should stick to the same categories that we use for ores, such as "is mountain", "is volcano", and so on. (They have a tag in core if you want the full list).

Anything which we want to be spawned in certain climates (temp, rainfall, or forest amount) should be put in as many biomes as possible, then controlled via tfc's fauna system (I think you wanted herbal ones to only spawn in forests?)

Also, will these spawn above ground too? I think players kinda like that the overworld's surface is mostly fantasy-free

SakuraKitsurugi:
I guess Rock and Phosphorum can spawn in caves, but Herbal would not really make sense and I do not think we have lush caves in overworld.

Pyritie:
hm, maybe herbal could grow in any forest as long as they were OLD_GROWTH? those are the densest ones that sometimes also spawn vanilla monsters because of how dark they get

Pyritie:
oh maybe phosphorum could grow in cold forests and herbal grow in warm forests?

SakuraKitsurugi:
Sounds good, for Rock I mostly had the idea that it only spawns in caves of old/ancient mountains and volcanoes.

Pyritie:
if that's the case I worry it'd be too rare for players to ever encounter

Pyritie (approved):

PR #549, opened 2026-08-04
Remove inconsistent heat based on block tags.

ashleney:
- fix inconsistent magma and lava temperature
- if the bug with a 10000C crucible happens, at least don't let it kill the player
- move the leather apron back to pants curio slot


Pyritie (approved):

SakuraKitsurugi (approved):

Pyritie (commented):
oh, there's some things in the modpack that still use `warm_stuff`

ashleney:
oh true, lemme add those

SakuraKitsurugi (approved):

Pyritie (approved):

PR #556, opened 2026-08-13
Shader uniforms

Pyritie:
not sure why these bud files are here, maybe something leaked over from my combustion engine branch


ashleney (approved):

src/main/java/su/terrafirmagreg/core/mixins/common/tfc/new_ow_wg/TFCBiomesMixin.java
Pyritie, line 1: not sure why this is here either. But it's "correct" I guess

PR #571, opened 2026-09-01
Me assembler

TomPlop:
-Remove OC from the ME Assembler and make Batch Mode native
-Adjust the formula and the damage curve (only duration increases damage)
-Save the HP in the Controller so players can't abuse the system
-Adjust some GUI


Pyritie (approved):

ashleney (commented):
What would happen if you swapped an exquisite certus quartz block to a basic certus quartz block in the middle of running a recipe? From what I understand the recipe will continue running and players could bypass the charger.

TomPlop:
> What would happen if you swapped an exquisite certus quartz block to a basic certus quartz block in the middle of running a recipe? From what I understand the recipe will continue running and players could bypass the charger.

No the damage are made on BeforeWorking which is when the recipe starts. So the recipe starts, your budding takes damage, you remove it then the recipe ends and no recipe starts again. You add it again, same health, it takes damage and the recipe starts.

There is one way to abuse the system but it's so fucking complex, I think it's not even worth it.

ashleney (approved):

PR #575, opened 2026-09-07
Jetpack buffs

Pyritie:
- New liquid fuel jetpack to replace the base GT one, because mixins are being tricky. Its hover is as good as the old advanced nano.
- Liquid fuel jetpack now also accepts all liquid blaze burner fuels for early accessibility (so things like creosote, lava, cooking oils, etc)
- Basic electric jetpack now also has the same hover as the advanced nano
- Advanced electric and advanced nano both now have perfect hover


TomPlop:
Didn't we say that we bump hover efficiency but drastically reduce speed? So other methods of transportations still have a reason to exist?

The Quadcopter was bad enough but I feel jetpack at ULV is even worst, especially with any fuels that nullifies everything else like jump boost/feather falling.

If you want to keep it like the Quadcopter at least maybe give it an awful speed and increase drastically the fuel consumption? I'm afraid right now it's just removing aspect of the game.

Pyritie:
the liquid one already has an awful speed, it's about as fast as a normal walk. This is also actually pushed further back than the quadrocopter since that one was just steel age, whereas this new one will be mid-steam age, around the time you get the create-themed car and plane

I can change the fuel consumption but it's already drinking it pretty fast, like 10mb every second

TomPlop:
The speed isn't that bad, with hover mode it's a bit faster that running or about the same speed as quadcopter while removing any cons of it. Without hover mode you go up and down at an insane speed.

The fuel consumption depends of the rate in the Blaze Burner so right now it may accept Steam and Olive Oil would be incredibly goated? A good consumption makes it useful only for building (which was the idea?) rather that removing part of the game that are already under used.

Also regarding the other jetpacks, I thought we said buff hover mode but nerf speed was the middle ground so they aren't the best transportation method?

TomPlop:
Anyway ultimately there is the new modular armor coming before the nano one with GTm8 so stuffs can be change then. If we want to allow some diversity rather than a solution that does it all.

Pyritie:
> The speed isn't that bad, with hover mode it's a bit faster that running or about the same speed as quadcopter while removing any cons of it. Without hover mode you go up and down at an insane speed.

it's really not that fast vertically either lol. But ok I can reduce the speed more, since this is mostly for people to build with

> The fuel consumption depends of the rate in the Blaze Burner so right now it may accept Steam and Olive Oil would be incredibly goated? A good consumption makes it useful only for building (which was the idea?) rather that removing part of the game that are already under used.

That's what the `/ 10` in the `registerFuel()` is for -- it takes the "olive oil burns in a blaze burner for 2.5 seconds" into 0.25 seconds per mB. If this isn't enough then we can make it /15 or /20

> Also regarding the other jetpacks, I thought we said buff hover mode but nerf speed was the middle ground so they aren't the best transportation method?

The electric ones still drain power really quickly so they're very bad for long distance travel. I can make them drain even faster if we want though, since we have the wireless charger to make up for it

TomPlop:
Yeah I think that would be good change reducing speed a bit more so you really feel the drain if you try to use it over other methods of transportation.

It's around an hour for a full tank of Olive Oil. Also with the way it works every fuels that consume less that 10mb per tick are being nerfed and stuffs like steam with their usually 600mb consumed per tick in the Blaze Burner are getting an insane buff. Basically using the burnTime of Blaze Burner makes it really hard to balance every fuels. By the way shouldn't we remove steam from it at that point?

If you really want to keep the speed for the electrical one, I would be very favorable to a massive nerf on energy consumption. Then obviously the Advanced one is just a shit show.

I'm sorry to be bothering on these changes, but I really want Immersive Vehicles to feel useful in the game which will be harder with how it's more difficult to fly with it. Also I'm still really sad no one is using all the drugs that exist and I would like them to keep have space to exist.

Pyritie:
> It's around an hour for a full tank of Olive Oil. Also with the way it works every fuels that consume less that 10mb per tick are being nerfed and stuffs like steam with their usually 600mb consumed per tick in the Blaze Burner are getting an insane buff. Basically using the burnTime of Blaze Burner makes it really hard to balance every fuels. By the way shouldn't we remove steam from it at that point?

We can, but the last time I tried that it was crashing peoples' worlds who had steam inside the blaze burner

> I'm sorry to be bothering on these changes, but I really want Immersive Vehicles to feel useful in the game which will be harder with how it's more difficult to fly with it. Also I'm still really sad no one is using all the drugs that exist and I would like them to keep have space to exist.

There's still a lot that the drugs can do that vehicles/jetpacks can't, and if we really want people to use them then we need to stop letting them just jump off a building to remove all of the medical conditions

src/main/java/su/terrafirmagreg/core/common/item/wearable/LiquidFuelJetpack.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class LiquidFuelJetpack extends PowerlessJetpack {
+    @Override
+    public double getVerticalHoverSlowSpeed() {
+        return 0.005D;
+    }
+
+    @Override
+    public double getVerticalHoverSpeed() {
+        return 0.15D;
```
Redeix, line 23: Might be nice for these to be config values tbh

Pyritie: augh

Pyritie: There's too many numbers involved and this will also open up the option for people to just make the starter jetpack as good as quarktech

PR #576, opened 2026-09-07
Ingot Pile Changes

Redeix:
- Merges the sheet and ingot pile mixins into the tfc classes.
- Merges pile renderer handlers into tfc.
- Adds the ability to hold the sprint key to add and subtract whole stacks of ingots at a time.
- Adds insert and extract methods to piles to allow things like hoppers and item pipes to move pile inventories.
      - Capped at 64 items a tick.

Currently no method to place intial piles automatically. But maybe ashlen or someone can bring back the ingot pile manager that they were working on.
Tested both runClient and modpack side.


Redeix:
oh, also works for double ingot piles

Pyritie (commented):
which part of this is new and which part is mixin refactors 🥀

Redeix:
> which part of this is new and which part is mixin refactors 🥀

lol, good question. pretty much everything new is in `IngotPileBlock` and `IngotPileBlockEntity` with a bit of new stuff in `InteractionManager`. I did clean up some of the other code to remove warnings and errors but they should all function the same.

Pyritie (approved):

PR #584, opened 2026-09-14
Fix crowbars not being extractable from create logistics, again

### LINKED ISSUE #4794: [🐛]: Red steel crowbars cannot be retrieved once they enter a Create storage container.

- [x] No, it was not.

### Computer OS

windows 11

### Modpack Version Used

0.13.7

### Minecraft Environment

Essential

### New Worlds

Unsure

https://mclo.gs/tyuDBT7

### Expected Behavior

When ordering Red Steel Crowbars via the Blaze Burner, a Red Steel Crowbar will be retrieved from the Create storage by the Packer and delivered to the crate.

### Actual Behavior

When attempting to order via the Flare Burner, the Crowbar remains in the interface and cannot be retrieved; retrieval via other methods is also impossible.

### Steps to Reproduce

1) Put the crowbar into the "create" storage.
2) Ordered a crowbar via the Flare Burner.
3) The crowbar remained in place.

### Additional Information

_No response_

ashleney:
so I looked more deeply into it and it turns out gt tool's nbt field `GT.Tool` is [dynamically updated on read](https://github.com/GregTechCEu/GregTech-Modern/blob/1.20.1/src/main/java/com/gregtechceu/gtceu/api/item/IGTTool.java#L304) and because of how create logistics happens to compare items the item will always report itself as not being equal. I think the reason why this happened specifically crowbars and not other tools is because people only ever used the crowbar for maintenance and never anything else. At least that's what I think.

the reason you couldn't extract sns sacks is because they send data only [in the share tag](https://github.com/traister101/Sacks-and-Such/blob/1.20/src/main/java/mod/traister101/sns/common/items/ContainerItem.java#L243-L251) but not in the main tag. 

I am not aware of any way to affect specifically create with this mixin but I don't think it should cause any issues.


Pyritie (approved):

Redeix (approved):
scariest if statement

ashleney:
I literally could not figure out a more elegant way to do it, trust me you don't want to see my prototypes

PR #586, opened 2026-09-16
building gadgets ftbchunks compat

ashleney:
Every time a block is placed/removed we check if the responsible player has EDIT_BLOCK permissions in that block pos. building gadgets queues all the blocks that are about to be placed/removed so that it only places only up to 50 blocks per tick which is why it happens to bypass ftbchunks.


Pyritie:
is this going to be firing every block event from everything? or just from building gadget stuff?

ashleney:
just building gadgets, here's what's being mixin'd: https://github.com/Direwolf20-MC/BuildingGadgets2/blob/1.20.1/src/main/java/com/direwolf20/buildinggadgets2/common/events/ServerTickHandler.java#L51-L60

Pyritie (approved):

TomPlop (approved):

PR #587, opened 2026-09-17
Fix TMRV crashing when exiting right after world join

Redeix:
Fixes `NullPointerException` crash when exiting right after world join before EMI finishes loading. Basically just caused by improper cache cleanup on player logout while initializing so I filled it with catches.
Tested in `runClient` and modpack.
Hecuba really hates this issue.


Xikaro:
Strangely, I don't have any similar problems.

Redeix:
I was able to recreate it. But for me it only happens within like half a second of joining a world. Hecuba states that it's a really annoying issue when speedrunning

Redeix:
I believe @Filiipa has experienced this a bunch before. Or so I am told.

Pyritie (approved):

ashleney (approved):
it's full of catches indeed

PR #588, opened 2026-09-18
lamp fuel recipe category

Redeix:
Lamp fuel tab was not included with the fork. So adding it back in.


Pyritie:
hm I thought we forked 3.2.25

PR #589, opened 2026-09-18
added hemisphere information to ClimateScreen when on overworld

### LINKED ISSUE #4145: [💡]: Add hemisphere indicator to Calendar / Climate sidebar tabs.
### Was this feature suggested before?

- [x] No, it was not.

### A list of popular features/mods the team has already considered and will not implement.

- [x] I understand :(

### What would you like to be added?

Title

### Why is this needed?

Reverses the climate entirely but there's no obvious indication why. This would help.

ariedotme:
## What is the new behavior?
Now, when on the overworld, there's a new label on the climate screen that tells what's the current hemisphere.

## Implementation Details
On this branch the hemisphere info is public, easily accesible and stuff so I basically just added the new label on the ClimateScreen and had to change all the label's positions by 3px so it fits nicely.

Screenshot 2026-09-18 at 17-39-16


Pyritie:
what happens if this was enabled on other planets? I know they don't have any noticeable differences between north and south, but would it actually work? they do have some differnent sizes, like I think mars is 3k from equator to pole instead of 10k

Pyritie (commented):
also, the langs need to be submitted here for our translation system: https://github.com/TerraFirmaGreg-Team/Tools-Modern/tree/dev/LanguageMerger/LanguageFiles/tfg/en_us

ariedotme:
Right now there are two guards

if (player == null || player.level().dimension() != Level.OVERWORLD)
if (!(Climate.model(player.level()) instanceof OverworldClimateModel model))

The dimension one is redundant, TFC only uses OverworldClimateModel in the overworld and everything else falls back to BiomeBasedClimateModel. So the instanceof already blocks every other planet on its own, and I think we can remove the dimension check.

The maths would already work, getInNorthernHemisphere takes the scale as a parameter and we pass model.getTemperatureScale(), so if Mars had its own OverworldClimateModel with 3k instead of 10k it would just scale, nothing to change, what's missing is the model not the calculation.

To do it properly I'd move the question onto ClimateModel, It's an interface with a bunch of default methods already, so it could have one that answers "no hemispheres here" (probably null) by default, with OverworldClimateModel overriding it, then the screen just asks the model without caring abou the dimension,  the Beneath would have no hemispheres and the line wouldn't render.

Also one thing I noticed is that Mars already has MartianPolesFeature, and the comment in it says most of the code came from OverworldClimateModel::onChunkLoad. So there's this logic on Mars already but just living outside the climate model system... I think it wouldn't be hard to make it work

Pyritie:
Ah ok! Nah it's fine if it's a longer bother. The `onChunkLoad` part is for it calculating snow noise I think, unrelated to anything about the climate model

Pyritie (approved):

Redeix (approved):

